PR #416, opened 2011-02-10
:hash interpolation for generic obfuscated URLs and :timestamp bugfix

jmileham:
Hi there,

I'm building a site that needs to securely obfuscate attachments in a public S3 bucket.  The solution I arrived at seems like it might be useful to the community as a general-purpose tool that remains customizable enough to allow developers with different requirements to implement it differently without having to hack into Paperclip core.  It is a set of storage-agnostic extensions to Attachment and Interpolations and supporting tests.

It features:
- No extra data model required
- Choice of HMAC digest algorithm on a per-attachment basis via openssl (modeled after ActiveSupport::MessageVerifier, default algo is SHA1)
- Choice of fields to include in the hash on a per-attachment basis, which permits many developer-selected tradeoffs between security and flexibility -- some examples:
  - `:hash_data => ":class/:attachment/:id"` - Unique per asset, long-lived filename with deducible style URLs (e.g. you want a 3rd party site to be able to generate style URLs from a single URL trunk that you provide once and always get the latest attachment)
  - `:hash_data => ":class/:attachment/:id/:style"` - Adds non-deducible style URLs (e.g. you want to store uncompressed :originals but don't want end users to be able to find them and kill your bandwidth bill)
  - `:hash_data => ":class/:attachment/:id/:style/:updated_at"` - Adds non-deducible versions (e.g. if a user's private profile photo URL makes it out into the wild via a malicious friend, they can unfriend, upload a new photo and the leak is plugged) -- this is the default

Since the hashes are extremely unlikely to collide, you can also remove other elements of the `:path` that were previously required for global uniqueness (like `:id`), which can give your users plausible deniability of ownership, or even completely obliterate the directory structure (`:path => ":hash"` -- or less controversially `:path => ":hash.:extension"`) such that an attacker can't even prove what kind of attachment he/she has gained access to.  Paired with a large `:hash_secret` and HTTPS, your users are left with a similar security guarantee to the analog hole -- attackers would need an authorized-user mole, or would need to compromise either your server-side secret or an authorized user's machine in order to gain access to protected attachments, and then would be able to freely distribute them (sadly, using your hosting infrastructure as an accomplice).
### The `:timestamp` bug

While building this, I ran across what I think is a bug: The `:timestamp` interpolation is non-deterministic in the presence of per-thread time_zones, e.g. displaying time zones per user location, which would affect any such site that uses `:timestamp` in an attachment :path or :url (or if I had used `:timestamp` in `:hash`).  While using `:timestamp` in paths and URLs is probably a little-used feature given the general wordiness of `Time#to_s`, and the intersection of sites that attempt to use it as well as implement per-user time zones is a vanishingly small slice of Paperclip users, I think it's worth taking a look at.

I took a stab at dealing with it by adding a new attachment config parameter `:use_default_time_zone` (defaulted to true) that explicitly churns out :timestamp interpolations in the server-wide default time zone.  This is not perfect, because it's possible that a site owner would want to change the default time zone via `config.time_zone=`, which would in turn invalidate all of their existing attachment URLs.  One option available to such an implementor would be to redefine my new method `Attachment#time_zone` explicitly to meet their requirements.  Or I could add a `:time_zone` option on `Attachment`, defaulted to null, and falling through to the present `Attachment#time_zone` implementation.  Such an option could even optionally take a block a la `options[:url]` and `options[:path]`.

While the ideal path for `:timestamp` is unclear to me as I'm not deeply familiar with the design philosophy behind Paperclip or the direction it wants to head in (aggressively fixing holes like this or supporting users who may rely on legacy behavior?), it seemed natural to me to expose the integer `Attachment#updated_at` value as an interpolation as an alternative to `:timestamp`.  Epoch seconds don't carry any time zone information and are therefor immune to this issue, plus they're presumably faster to convert to than text, and fewer characters to hash to boot.

So my present implementation has a provisionally "improved" behavior for :timestamp (which should be backward compatible for everybody who didn't run into the bug personally), and a new `:updated_at` interpolation that merely exposes the existing `Attachment` instance method.  I'd be eager for feedback on what if anything should be done with that stuff.

The caveats:
- I'm pretty new to testing, and this was definitely my first time writing tests using shoulda and test/unit.  Feel free to slap me for transgressions in style and abuse of mocks/stubs/fakes/dummies.
- I tried to follow house style to the extent I could derive it, but I'm sure I got it wrong in places.
- I'm not sure what went on with the appraisal gemfile.locks reverting to an earlier version dependency on Rails -- perhaps this is expected?  Again, never used appraisal before either.  Tips would be more than welcome.

I'd be very excited if you see some or all of these changes as valuable to the Paperclip mainstream, and am happy to iterate this into something worthy of pulling.

Thanks a lot!

-john

### DIFF
--- a/gemfiles/rails3.gemfile
+++ b/gemfiles/rails3.gemfile
@@ -2,7 +2,7 @@
 
 source "http://rubygems.org"
 gem "ruby-debug"
-gem "rails", ">=3.0.3"
+gem "rails", "~>3.0.0"
 gem "rake"
 gem "sqlite3-ruby", "~>1.3.0"
 gem "shoulda"
--- a/gemfiles/rails3.gemfile.lock
+++ b/gemfiles/rails3.gemfile.lock
@@ -90,7 +90,7 @@ DEPENDENCIES
   appraisal
   aws-s3
   mocha
-  rails (>= 3.0.3)
+  rails (~> 3.0.0)
   rake
   ruby-debug
   shoulda
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -8,16 +8,19 @@ module Paperclip
 
     def self.default_options
       @default_options ||= {
-        :url               => "/system/:attachment/:id/:style/:filename",
-        :path              => ":rails_root/public:url",
-        :styles            => {},
-        :processors        => [:thumbnail],
-        :convert_options   => {},
-        :default_url       => "/:attachment/:style/missing.png",
-        :default_style     => :original,
-        :storage           => :filesystem,
-        :use_timestamp     => true,
-        :whiny             => Paperclip.options[:whiny] || Paperclip.options[:whiny_thumbnails]
+        :url                   => "/system/:attachment/:id/:style/:filename",
+        :path                  => ":rails_root/public:url",
+        :styles                => {},
+        :processors            => [:thumbnail],
+        :convert_options       => {},
+        :default_url           => "/:attachment/:style/missing.png",
+        :default_style         => :original,
+        :storage               => :filesystem,
+        :use_timestamp         => true,
+        :whiny                 => Paperclip.options[:whiny] || Paperclip.options[:whiny_thumbnails],
+        :use_default_time_zone => true,
+        :hash_digest           => "SHA1",
+        :hash_data             => ":class/:attachment/:id/:style/:updated_at"
       }
     end
 
@@ -32,24 +35,28 @@ module Paperclip
 
       options = self.class.default_options.merge(options)
 
-      @url               = options[:url]
-      @url               = @url.call(self) if @url.is_a?(Proc)
-      @path              = options[:path]
-      @path              = @path.call(self) if @path.is_a?(Proc)
-      @styles            = options[:styles]
-      @normalized_styles = nil
-      @default_url       = options[:default_url]
-      @default_style     = options[:default_style]
-      @storage           = options[:storage]
-      @use_timestamp     = options[:use_timestamp]
-      @whiny             = options[:whiny_thumbnails] || options[:whiny]
-      @convert_options   = options[:convert_options]
-      @processors        = options[:processors]
-      @options           = options
-      @queued_for_delete = []
-      @queued_for_write  = {}
-      @errors            = {}
-      @dirty             = false
+      @url                   = options[:url]
+      @url                   = @url.call(self) if @url.is_a?(Proc)
+      @path                  = options[:path]
+      @path                  = @path.call(self) if @path.is_a?(Proc)
+      @styles                = options[:styles]
+      @normalized_styles     = nil
+      @default_url           = options[:default_url]
+      @default_style         = options[:default_style]
+      @storage               = options[:storage]
+      @use_timestamp         = options[:use_timestamp]
+      @whiny                 = options[:whiny_thumbnails] || options[:whiny]
+      @use_default_time_zone = options[:use_default_time_zone]
+      @hash_digest           = options[:hash_digest]
+      @hash_data             = options[:hash_data]
+      @hash_secret           = options[:hash_secret]
+      @convert_options       = options[:convert_options]
+      @processors            = options[:processors]
+      @options               = options
+      @queued_for_delete     = []
+      @queued_for_write      = {}
+      @errors                = {}
+      @dirty                 = false
 
       initialize_storage
     end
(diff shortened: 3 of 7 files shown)


quentindemetz:
## I am having a problem with this feature and am losing my images.

**Consider the following scenario for a likely explanation:**
I have in my initializer:
```ruby
Paperclip::Attachment.default_options[:hash_data]    = ':class/:attachment/:id/:style/:updated_at'
```

and my `Person` model:
```ruby
class Person < ApplicationRecord
  has_attached_file :avatar,
      styles: { tiny: ['70x70#', :jpg] },
      hash_secret: ENV['PAPERCLIP_HASH_SECRET'], 
  [...]
end
```

Everything works fine! I'm happy, my image urls are obfuscated, looks cool and shiny.

Tomorrow I add a new style:
```ruby
styles: { tiny: ['70x70#', :jpg], medium: ['100x100#', :jpg }
```

I follow the paperclip documentation to process all my existing images for the new `medium` style. I create the relevant .yml file with the following contents:
```
---
:Person:
  :avatar:
    - :tiny
```

And I run the following rake task (defined in https://github.com/thoughtbot/paperclip/blob/master/lib/tasks/paperclip.rake)
```
rake paperclip:refresh:missing_styles
```

According to the rake task definition, it only generates the styles that are missing.
In my case, it generates only the ':medium' style by calling `person.avatar.reprocess!(:medium)`

However, this operation in turn updates the `avatar_updated_at` column. This "invalidates" all my previous images (including the original one) as the hash data has changed but the images have not been renamed. Life sucks.

jmileham:
Howdy, just FYI, this feature predates `refresh:missing_styles` by 6 months and at the time of its addition to Paperclip, the best practice (and in fact only way to get new styles rendered) would have been to refresh all the styles when new style definitions were added, which would work fine. Anyway, as described, it does sound like a gnarly bug and should definitely be addressed.

While it's tempting to include a tldr with simple fix that seems like a no-brainer (why was it even designed this way to begin with?), there's a tradeoff here that's worth acknowledging. What you suggest might be the right move all-in, but it's worth noting that when you remove updated_at from the mix, the filename will not change if the file is replaced with a new file.

For example, if you're on a social network with profile photos backed by the hash feature and a stalker manages to get the URL of your profile photo (say through a one-time compromise of a laptop of a friend), the stalker will forever be able to get your latest profile photo. That's a big difference in contract, and probably a surprising one to an app developer.

Another alternative that wouldn't change the security properties here would be simply to disable the `refresh:missing_styles` rake task when using hash_data that includes avatar_updated_at, or possibly suppress touching avatar_updated_at when refreshing missing styles (not sure this is possible or desirable, but it might actually be more semantically correct).

Or a third path might be to add an optional self-incrementing file version number column to the hash that indicates that a new file has been uploaded. It would provide you with the new secret URL for each distinct file uploaded. It would keep from unnecessarily changing the hash during operations that are fundamentally no-ops on the avatar itself. This would require the app developer to carry around more columns in their schema, though, which seems painful.

I'm sure the debate and a pull request would be welcome to help fix it. Sadly I think these two features came in from completely different contributors who weren't aware of the potential interplay. Hashing, being a lossy operation, can end up in these kinds of knife-edge situations, and it's too bad that an unrelated feature ended up having such a scary edge case.

jmileham:
A quick dive reveals that [refresh:missing_styles](https://github.com/thoughtbot/paperclip/blob/093b4da18444d883de826f58628089c2f480d461/lib/tasks/paperclip.rake#L89) calls [refresh:thumbnails](https://github.com/thoughtbot/paperclip/blob/093b4da18444d883de826f58628089c2f480d461/lib/tasks/paperclip.rake#L48) which calls [reprocess!](https://github.com/thoughtbot/paperclip/blob/7edb35a2a9a80c9598dfde235c7e593c023fc914/lib/paperclip/attachment.rb#L344) which calls [assign](https://github.com/thoughtbot/paperclip/blob/7edb35a2a9a80c9598dfde235c7e593c023fc914/lib/paperclip/attachment.rb#L112) which calls [assign_attributes](https://github.com/thoughtbot/paperclip/blob/7edb35a2a9a80c9598dfde235c7e593c023fc914/lib/paperclip/attachment.rb#L438) which calls [assign_timestamps](https://github.com/thoughtbot/paperclip/blob/7edb35a2a9a80c9598dfde235c7e593c023fc914/lib/paperclip/attachment.rb#L458) which blows away updated_at, which is only a problem if you're not updating all the styles at once.

It seems like reprocessing isn't really an assignment and probably shouldn't be exercising that complete codepath unmodified. That'd by my recommendation for resolution.

NeoElit:
I'm sorry to comment on this closed issue. But isn't this issue worth mentioning in the Readme? that [refresh:missing_styles](https://github.com/thoughtbot/paperclip/blob/093b4da18444d883de826f58628089c2f480d461/lib/tasks/paperclip.rake#L89) will break the urls if updated_at is included in the hash_data which is the case by default

jmileham:
I'd be in favor of some resolution, for sure. A readme comment seems like potentially admitting defeat rather than making the behavior less surprising, but could be a good stopgap. @jyurek any thoughts?

fwoelm:
Would a simple fix be to just "skip" updating `updated_at` when refresh:missing_styles is called? Given that `updated_at` should ideally tell us the last time the **user** updated their profile picture/.../..., we might not want `updated_at` to change anyway!

I do not understand the code layout well enough to check if this would be an easy fix or not. Maybe one of you who is better informed, can have a look :)

You would want to use (Rails 5)

     attachment.save(:touch => false)

or (Rails pre-5, I think)

     attachment.save(:updated_at => attachment.updated_at)

See: http://blog.bigbinary.com/2016/05/09/rails-5-allows-updating-without-updating-timestamps.html

fwoelm:
Or what if the value for :hash_data defaulted to `:class/:attachment/:id/:style/:file_size`. There's a chance that you'll have exactly the same file size for two images -- but if I'm not mistaken, the odds are rather low?

PR #916, opened 2012-06-16
Implemented performance improvement to interpolator

phene:
This patch introduces a performance improvement and some minor behavior changes to the interpolator.

Results of benchmarking have shown:

``` ruby
[":class/:attachment/:style/:id/:filename", ":class/:id/:filename"].each do |path|
  should "have performant path #{path}" do
    ObjectSpace.garbage_collect
    rebuild_model :path => path
    @dummy = Dummy.new
    @dummy.stubs(:id).returns(1024)
    @file = File.new(fixture_file("5k.png"), 'rb')
    @dummy.avatar = @file

    puts (1..3).map { Benchmark.realtime { 1000.times { @dummy.avatar.path } } }
    @file.close
  end
end
```

```
Interpolate ":class/:attachment/:style/:id/:filename"
  before
    0.488147
    0.488169
    0.491802
  after
    0.326616
    0.327032
    0.305505
  improvement
    34.7 %

Interpolate ":class/:id/:filename"
  before
    0.366626
    0.362184
    0.361812
  after
    0.209302
    0.184409
    0.210002
  improvement
    44.6 %
```

Additional thanks to @seanwalbran for his original implementation

### DIFF
--- a/test/attachment_test.rb
+++ b/test/attachment_test.rb
@@ -208,7 +208,7 @@ class AttachmentTest < Test::Unit::TestCase
 
     teardown { @file.close }
 
-    should "make sure that they are interpolated correctly as long non-word characters separate them" do
+    should "make sure that they are interpolated correctly as long as non-word characters separate them" do
       assert_equal "1024.omg/1024-bbq/:idwhat/000/001/024.wtf", @dummy.avatar.path
     end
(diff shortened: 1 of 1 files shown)


ideaoforder:
+1

Issue #1086, opened 2012-11-14
Support for Rails4


tobidelius:
It seems to be some problem with paperclip and rails master (rails.4.0.0.beta) at the moment. I get the following error when I try to upload something.

`wrong number of arguments (2 for 1)`

The exact same code works in rails 3.2.8.

PR #1106, opened 2012-12-10
Fix use of run_callbacks in Rails 4.

jviney:
I had same problem as #1086.

This fixed it for me.

### DIFF
--- a/lib/paperclip/callbacks.rb
+++ b/lib/paperclip/callbacks.rb
@@ -22,8 +22,8 @@ module Paperclip
     end
 
     module Running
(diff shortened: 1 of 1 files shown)


taimoor:
In case you are upgrading your rails version from 3 to 4. Then you need to upgrade your paperclip version too.

I was using `paperclip 3.0.4` in `rails 3.2.21`, on update to `rails 4.0.4` I've to update my Gemfile 

`gem paperclip "~>4.2.1"

gem 'cocaine', '0.5.3'`

PR #1346, opened 2013-09-21
Add mongoid support

johnvuko:
I wrote few lines for have mongoid support in paperclip, really needed. I have to change some little things for be able to do all tests already written for active_record with mongoid (added some if).
Just have to include `Paperclip::Glue` in the model for use with mongoid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,24 @@ gemfile:
   - gemfiles/3.1.gemfile
   - gemfiles/3.2.gemfile
 
+env:
+  - ORM=active_record
+  - ORM=mongoid
+
 matrix:
   allow_failures:
     - rvm: jruby-19mode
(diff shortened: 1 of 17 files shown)


dfabreguette-ap:
Hi guys, can we hope to have any native support on mongoid anytime soonish ? Or is it something utopic ? We're currently using this plugin https://github.com/meskyanichi/mongoid-paperclip, is this a good idea ?

yordis:
@jyurek is there any possibility of see this happening. 2016, almost, a lot of people are writing the new web apps using MongoDB (I have 1 year without touch SQL DB).

Please change the roadmap. I know it's not a easy path but this is something that everyone needs in these days. I am using mongoid-paperclip and so far it works but I would like to see just one gem for that, please.

+1

PR #1519, opened 2014-04-21
Assigning a file sets the correct content_type

quainjn:
This partially reverts https://github.com/thoughtbot/paperclip/commit/41c092d9d8801ac88fbfdfff8dfdc9262444cc86.  

When a CSV file is created assigned as the attachment, the mime type detected by the command line is "text/plain", not "text/csv".


jasonswett:
What's the word on this one? This appears to address the exact issue I'm having due to `file` apparently not always returning the right MIME type.

(Here's a Stack Exchange question that describes what I'm talking about when I say `file` apparently doesn't always return the right MIME type: http://unix.stackexchange.com/questions/185216/file-command-apparently-returning-wrong-mime-type)

jasonswett:
I'm also confused about how this problem hasn't been experienced more widely.

maclover7:
Hi @quainjn ! Is this still an issue? If so, can you please rebase your PR and make sure everything is passing? Thanks!

tute:
We need to check the mime types through `mimemagic` or the `file` command for security reasons. We can't merge this PR as is. Thanks for your input and sharing your work! :)

Issue #1506, opened 2014-04-06
Discussion about content type detection


mmustala:
I find it a bit confusing how content type validations are handled currently. Current way is to use the operating systems file command to detect the content type, which forces us to write OS dependent code.

For example I have an application that receives email messages and attachments. I have a MailReceiver class that picks some attachments from the emails based on their content type using a code

    message.attachments.each do |attachment|
      if attachment.content_type.match('application/vnd.ms-excel')
        save_attachment(attachment)
      end
    end

save_attachment saves the attachment as a paperclip attachment. Then paperclip does its own content_type validation and originally I had there only:

    validates_attachment_content_type :file, :content_type => %w(application/vnd.ms-excel)

And that worked for a while. Then I deployed my application to a Debian server and surprised that I was getting content_type validation errors. Errors came because my servers file command returns 'application/msword' for an Excel file. So I needed to add that into allowed content types in paperclip validation. 

Notice that I have not needed to adjust the content_type check in the MailReceiver. It still uses 'application/vnd.ms-excel'. I have not studied how Mail gem is determining the content_type but it apparently uses a different strategy for it.

I'm just wondering if I some day change server and I'm again getting weird errors because of a file command returning something else. It looks like it is completely in the hands of the people who write these packages. 

I don't like my application to be so coupled with the server and hope that these content type checks could be done using Ruby code. Even if that means adding another gem dependency. At least then I could be sure that Paperclip behaves the same on every server where I deploy my application.

maclover7:
Is this fixed with #1839 ?

tute:
Ping @mmustala: is this fixed now?

jyurek:
@mmustala Also there is now the option to turn off spoof checks, with the `validate_media_type: false` option.

tute:
> Also there is now the option to turn off spoof checks, with the validate_media_type: false option.

Is that unsafe, or is it sometimes ok? If it is unsafe, we shouldn't provide the option?

jyurek:
The intent is to be "secure by default", but since the option gets in the way of some people using it, as we see here, there's an option to turn it off.

tute:
@mmustala paperclip now uses mimemagick, which might help with the problem you were having.

Since it's been two years since your last activity on this issue, I hope it's working well now for you. If it's not, let us know and we'll continue the discussion. Thanks!

mmustala:
[...]

PR #1537, opened 2014-05-05
Fix mime checking when unix util "file" version <5.

jcoleman:
This is potentially relevant in #1506, but it doesn't really affect the current choices, rather, it merely ensures that the current code works even on older versions of the `file` command on unix/linux.

For versions prior to 5, the `-i` option provides the mime type. That option doesn't exist in version 5 and has been replaced with the `--mime` that Paperclip was using by default.

### DIFF
--- a/lib/paperclip/file_command_content_type_detector.rb
+++ b/lib/paperclip/file_command_content_type_detector.rb
@@ -12,10 +12,17 @@ module Paperclip
 
     private
 
+    def major_version_of_file_command
+      output = /^file-(\d+)\./.match(Paperclip.run("file", "-v 2>&1", {}, expected_outcodes: [0,1]))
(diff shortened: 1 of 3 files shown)


maclover7:

rubyisbeautiful:
@maclover7 It's no longer an issue for me/us, @cnk's findings are true for stock 5.10/5.11.  I assume it was something weird with our environment on that project (wouldn't be the first time).  Thanks!!

maclover7:
@jferris @jyurek Please close issue, problem appears to be solved.

jcoleman:
@maclover7 While I'm not personally affected by this anymore as we've moved to AWS (and a newer OS in the process), the code still has the same problem.

PR #1554, opened 2014-05-26
Duplicate error messages

kashif-umair:
This fix can be better explained in code so I'm commenting the issue in this PR. Here is the link for the comment.
https://github.com/thoughtbot/paperclip/pull/1418/files#r8854478

More discussion can be found here.
https://github.com/thoughtbot/paperclip/pull/1357/files#r6825945

I had sent another PR for the same issue but I was messed up with my old repository so I had to delete the older repository and the older PR #1418. Consider looking at #1418 also.


maclover7:

jyurek:
I agree with not having the error duplicated because not everyone uses form builders, but I would much rather see a "this is where we put the errors" option rather than just cut out a group of people without acknowledging their needs.

Please address this before this PR will be considered for merging. Thank you.

leob:
@rubiety, your approach ("after_validation :clean_paperclip_errors") is exactly what I am doing and for the time being it's adequate for me. Have to say I was a bit surprised it works this way though.

yordis:
:+1:

Extrapolator214:
I just added an initializer to override it

```ruby
# override in order to avoid duplicate error messages
module Paperclip
  module Validators
    class AttachmentContentTypeValidator
      def validate_each(record, attribute, value)
        base_attribute = attribute.to_sym
        attribute = "#{attribute}_content_type".to_sym
        value = record.send :read_attribute_for_validation, attribute

        return if (value.nil? && options[:allow_nil]) || (value.blank? && options[:allow_blank])

        validate_whitelist(record, attribute, value)
        validate_blacklist(record, attribute, value)

        if record.errors.include? attribute
          record.errors[attribute].each do |error|
            record.errors.add base_attribute, error
            record.errors.delete attribute
          end
        end
      end
    end
  end
end
```

kashif-umair:
I've been busy in my office work so much that I almost forgot this pull request and there has been a mishap, I accidentally deleted my paperclip repo fork. So now I cannot make any changes into this PR. Does anybody know about how to make changes to this PR? Well I'll try to open a new PR if anybody still needs this.

Extrapolator214:
I think if you create a new fork and branch with exactly the same name, when you push to it, it should update the PR. Not sure though, but it should work.

yordis:
:+1:

tute:
Rather than removing this, I'd workaround it as suggested by @rubiety:

```ruby
class Attachment < ActiveRecord::Base
  after_validation :clean_up_paperclip_errors

  private

  # This is because paperclip duplicates error messages
  # See: https://github.com/thoughtbot/paperclip/pull/1554 and
  # https://github.com/thoughtbot/paperclip/commit/2aeb491fa79df886a39c35911603fad053a201c0
  def clean_up_paperclip_errors
    errors.delete(:file)
  end
end
```

Or do `errors.delete(:file_content_type)` instead, depending on where you want the error to be retained.

---

Is this a good option? If so, should we document it in the README or wiki? Thanks!

tute:
https://github.com/thoughtbot/paperclip/pull/2090 fixes the issue by adding a configuration option.

tute:
This PR drops support for some form builders, and we don't want to drop support for that.

If you need to change this behavior, the workaround outlined in https://github.com/thoughtbot/paperclip/pull/1554#issuecomment-200666161 pleases both kind of users , and is simple enough to not deserve a generic fix in paperclip.

We won't merge this PR, and I'm willing to consider one that, without adding much complexity, takes care of both use cases.

Thank you very much for your time discussing this issue and suggesting implementations. :)

PR #1666, opened 2014-10-01
Adding -r option to file command

peterpengnz:
Adding `-r, --raw` option: Don't translate unprintable characters to \ooo.	
Normally file translates unprintable	characters to their octal representation.

without `-r` option, the file command some times return file type with `\00` like characters.

For example: 
On CentOS 5 and file command version 4.17
Test the mime type for a aiff audio file.

```
 file -b --mime mono_normalized.aiff
 audio/x-aiff\011
```

with `-r` option:

```
 file -b --mime mono_normalized.aiff
 audio/x-aiff
```

### DIFF
--- a/lib/paperclip/file_command_content_type_detector.rb
+++ b/lib/paperclip/file_command_content_type_detector.rb
@@ -15,7 +15,7 @@ module Paperclip
     def type_from_file_command
       type = begin
         # On BSDs, `file` doesn't give a result code of 1 if the file doesn't exist.
-        Paperclip.run("file", "-b --mime :file", :file => @filename)
+        Paperclip.run("file", "-b -r --mime :file", :file => @filename)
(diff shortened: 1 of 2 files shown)


maclover7:
Hi @peterpengnz ! Is this still an issue? If so, can you please rebase your PR and make sure everything is passing? Thanks!

peterpengnz:
Hi Jon

You could close/remove the request now. I had issues to replicate same
issue on other computers.

That's why write test cases are difficult. It may or may not pass on
different computers.

I guess if other programers experienced the extra tailing character issue,
they could delete the extra characters by using simple ruby code.

It won't be a big problem, I think.

Thank You

maclover7:
@jferris @jyurek Please close issue, problem appears to be solved.

PR #1700, opened 2014-11-15
Fast to slowly spec

soramugi:
```
 paperclip $ rspec spec/paperclip/integration_spec.rb --profile 1
Testing against version 4.1.7
.........................[deprecated] I18n.enforce_available_locales will default to true in the future. If you really want to skip validation of your locale you can set I18n.enforce_available_locales = false to avoid this message.
.........

Finished in 15.19 seconds (files took 0.60155 seconds to load)
34 examples, 0 failures

Top 1 slowest examples (9.54 seconds, 62.8% of total time):
  Paperclip Many models at once does not exceed the open file limit
    9.54 seconds ./spec/paperclip/integration_spec.rb:18
```

to

```
 paperclip $ rspec spec/paperclip/integration_spec.rb --profile 1
Testing against version 4.1.7
.........................[deprecated] I18n.enforce_available_locales will default to true in the future. If you really want to skip validation of your locale you can set I18n.enforce_available_locales = false to avoid this message.
.........

Finished in 10.95 seconds (files took 0.59585 seconds to load)
34 examples, 0 failures

Top 1 slowest examples (5.33 seconds, 48.7% of total time):
  Paperclip Many models at once does not exceed the open file limit
    5.33 seconds ./spec/paperclip/integration_spec.rb:18
```

at trim indent


maclover7:
Hi @soramugi ! Is this still an issue? If so, can you please rebase your PR and make sure everything is passing? Thanks!

soramugi:
@jyurek I wonder if had not been through the test when I think I had been once merge . But what with this because this time can be confirmed that the test has been through .

tute:
Thanks @soramugi! I rebased your history into this commit: https://github.com/thoughtbot/paperclip/commit/c989efed4cf3ad984f017319599a0b89f829283d.

But before I merge, can you please tell me why do we need the new gem? And also, is the spec still testing what it has to test, even after your division in 100 chunks?

Thank you!

tute:
> But before I merge, can you please tell me why do we need the new gem? And also, is the spec still testing what it has to test, even after your division in 100 chunks?

@soramugi we are close to merge! Please don't forget those questions! :+1:

tute:
Quoting from the project's README to answer my question:

> Say you had a schema like this: `Publishers have Books Books have Reviews` and you wanted to bulk insert 100 new publishers with 10K books and 3 reviews per book. This library will follow the associations down and generate only 3 SQL insert statements - one for the publishers, one for the books, and one for the reviews.
>
> In contrast, the standard ActiveRecord save would generate 100 insert statements for the publishers, then it would visit each publisher and save all the books: 100 * 10,000 = 1,000,000 SQL insert statements and then the reviews: 100 * 10,000 * 3 = 3M SQL insert statements,
>
> That would be about 4M SQL insert statements vs 3, which results in vastly improved performance. In our case, it converted an 18 hour batch process to <2 hrs.

tute:
Thank you! :clap:

PR #1708, opened 2014-11-22
Change MD5 Fingerprints to SHA1 Fingerprints

medcat:
MD5 can collide.  SHA1 is better.

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -33,7 +33,7 @@ module Paperclip
         :use_timestamp         => true,
         :whiny                 => Paperclip.options[:whiny] || Paperclip.options[:whiny_thumbnails],
         :validate_media_type   => true,
-        :adapter_options       => { :hash_digest => Digest::SHA1 },
+        :adapter_options       => { :hash_digest => Digest::MD5 },
         :check_validity_before_processing => true
       }
     end
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -33,6 +33,7 @@ module Paperclip
         :use_timestamp         => true,
         :whiny                 => Paperclip.options[:whiny] || Paperclip.options[:whiny_thumbnails],
         :validate_media_type   => true,
+        :adapter_options       => { :hash_digest => Digest::SHA1 },
         :check_validity_before_processing => true
       }
     end
@@ -96,7 +97,8 @@ module Paperclip
     # attachment:
     #   new_user.avatar = old_user.avatar
     def assign(uploaded_file)
-      @file = Paperclip.io_adapters.for(uploaded_file)
+      @file = Paperclip.io_adapters.for(uploaded_file,
+                                        @options[:adapter_options])
       ensure_required_accessors!
       ensure_required_validations!
 
@@ -519,15 +521,19 @@ module Paperclip
       begin
         raise RuntimeError.new("Style #{name} has no processors defined.") if style.processors.blank?
         intermediate_files = []
+        original = @queued_for_write[:original]
 
-        @queued_for_write[name] = style.processors.inject(@queued_for_write[:original]) do |file, processor|
+        @queued_for_write[name] = style.processors.
+                                  inject(original) do |file, processor|
           file = Paperclip.processor(processor).make(file, style.processor_options, self)
           intermediate_files << file unless file == @queued_for_write[:original]
           file
         end
 
         unadapted_file = @queued_for_write[name]
-        @queued_for_write[name] = Paperclip.io_adapters.for(@queued_for_write[name])
+        @queued_for_write[name] = Paperclip.io_adapters.
+                                  for(@queued_for_write[name],
+                                      @options[:adapter_options])
         unadapted_file.close if unadapted_file.respond_to?(:close)
         @queued_for_write[name]
       rescue Paperclip::Errors::NotIdentifiedByImageMagickError => e
--- a/README.md
+++ b/README.md
@@ -728,10 +728,10 @@ is specified in `:hash_data`. The default value for `:hash_data` is `":class/:at
 
 For more on this feature, read [the author's own explanation](https://github.com/thoughtbot/paperclip/pull/416)
 
-MD5 Checksum / Fingerprint
+Checksum / Fingerprint
 -------
 
-An MD5 checksum of the original file assigned will be placed in the model if it
+A  checksum of the original file assigned will be placed in the model if it
 has an attribute named fingerprint.  Following the user model migration example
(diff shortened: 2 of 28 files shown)


maclover7:

medcat:
Yeah, I'll rebase this, give me a little while.

medcat:
Updated, but I'm not sure what's going on with travis...

maclover7:
This looks set to go.

paging @jferris @jyurek

medcat:
Any progress on the status of this?

maclover7:
Hi @medcat! Currently just waiting on @jyurek or @jferris to say if this is ready to be merged into Paperclip. I think it's ready, but we need to wait for one of them (I don't have write permissions to the repo).

clemensg:
Hi! What about existing projects who update to this version.. shouldn't there be a way to automatically migrate their MD5 fingerprints to SHA1? A rake task to force the regeneration of all fingerprints would be nice.

maclover7:
paging @jferris @jyurek @mike-burns

jyurek:
This isn't going to get merged without a way to either automatically update the hashes or a way to switch between which class is doing the digesting. As I said before, there are a LOT of people already using MD5 hashes and this would break every single one. Please address this or it won't get merged. Thanks.

spec/support/matchers/have_column.rb
```diff
@@ -8,7 +8,7 @@
     column && column.default.to_s == @default.to_s
   end
 
-  failure_message_for_should do |columns|
+  failure_message do |columns|
```
jyurek: This is also in #1727, and i have the same question: Is this backwards compatible with RSpec 2? If not, can you remove this and reference it separately, as it's out of scope for this PR?

medcat: You're right, I forgot to take it out of the changes; it was getting annoying to run the rspec tests and having deprecation warnings :cold_sweat: 

However, looking through RSpec 3 and Rspec 2.14, it seems that it is not backwards compatible.

- [RSpec 3](https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers/dsl.rb#L123-L125): The method `failure_message` is defined (with `failure_message_for_should` defined much later)
- [RSpec 2.14](https://github.com/rspec/rspec-expectations/blob/v2.14.5/lib/rspec/matchers/dsl.rb): It instantizes a [`RSpec::Matchers::DSL::Matcher`](https://github.com/rspec/rspec-expectations/blob/v2.14.5/lib/rspec/matchers/matcher.rb#L139-L141), where `failure_message_for_should` is defined, but no `failure_message`.

Therefore, it should not be backwards compatible.

medcat:
Alright, that should be all of the changes.

Hound is frustrating...

medcat:
If you like, I can attempt to rebase it on to the master.

maclover7:
Please rebase on master. @jyurek @mike-burns Status here?

clemensg:
Great work @medcat !  :+1:

medcat:
I derped.

medcat:
This has been rebased to the master, again.  It can be merged at any time.

clemensg:
What's the status here on this PR? @jyurek @mike-burns @maclover7 @medcat

medcat:
There's one more modification I want to make before I'm going to be happy with it, give me a little bit to get it working

clemensg:
OK.

Some thoughts regarding MD5 being insecure and switching to SHA-1:
Maybe we should just use SHA-2 with 256 bits instead of SHA-1, otherwise we'll be switching again in a few years? It's already in sight that SHA-1 is unusable, especially since http://2012.sharcs.org/slides/stevens.pdf and we can expect it to become worse every year.

So I would suggest using SHA2 with 256 bit or maybe even SHA3 (Keccak) with 256 bit ? But SHA2 would definitely be much faster on current hardware as it is already hardware accelerated in most processors.

What do you think?

medcat:
I chose SHA1 because it was still decently short, and was still more secure than MD5.  However, if there are questions on the security of SHA1, I can (and will) move the default to SHA256/SHA384/SHA512.

On another note, a side effect of this pull request allows developers to choose the digest that they want; so, the more security conscious can go with a more secure hashing mechanism.

medcat:
I believe I am content with it.

lib/paperclip/attachment.rb
```diff
@@ -32,6 +32,7 @@ def self.default_options
         :use_timestamp         => true,
         :whiny                 => Paperclip.options[:whiny] || Paperclip.options[:whiny_thumbnails],
         :validate_media_type   => true,
+        :adapter_options       => { :hash_digest => Digest::SHA1 },
```
jyurek: I'm sorry to have to repeat this, but SHA1 cannot be the default. At least, not until we have a major version bump. This should remain MD5, and changing it to SHA1 (or anything else provided by `Digest`) should be well documented. There can (and should) still be tests for SHA1, but it should be specified as an option in those tests, not assumed that it's the default.

medcat:
It defaults to MD5 now.

bdewater:
I believe defaulting to MD5 is a mistake with this change possibly getting released with Paperclip 5.0 (#1991), where you can 'get away' with breaking changes. Keep it around as an option for existing users to add so they can upgrade the hashes at their own pace, but keep it secure by default. Go with SHA2, based on https://gist.github.com/tqbf/be58d2d39690c3b366ad

clemensg:
I second that. SHA-256 seems to be a safe default for Paperclip 5.0.
The upgrade rake task makes the update process easy.

I think it is a mistake to switch to SHA1 which has several known flaws. If we update then let's go straight to SHA2 (with 256 bits). SHA2-512 and SHA3 are probably overkill.

This PR is sitting around for some time now. Any remaining objections from the maintainers?

@jyurek @tute @mike-burns

tute:
Deferring to @jyurek.

tute:
I agree that if we change the hashing algorithm, it should be `SHA2` by default, and it should be configurable so users upgrading can continue to use `MD5` until needed. We can make this decision until we release the next major version, which is soon upcoming.

We want to comply with the coding standards and Hound comments for changed code, but leave alone the preexisting code. That way the changeset is smaller and easier to review. If there are only stylistic changes and you can undo them, please do. Thanks!

Can you please rebase on top of latest master? Thank you all for your input and work.

bdewater:
Unless somebody beats me to it, I'm willing to work the next week on rebasing this PR with SHA256 as default, incorporating the feedback from Hound and the above comment from @tute, so this can ship in Paperclip 5.0.

tute:
Closing in favor of @bdewater's https://github.com/thoughtbot/paperclip/pull/2229. Thank you.

PR #1722, opened 2014-12-12
add support for Proc in S3 storage for s3_options

Migoo:
Hello,

I add this support to use different S3 endpoints if buckets need it.

Thanks for adding it in the main gem.

Cheers
Migoo

### DIFF
--- a/lib/paperclip/storage/s3.rb
+++ b/lib/paperclip/storage/s3.rb
@@ -136,6 +136,8 @@ module Paperclip
 
         base.instance_eval do
           @s3_options     = @options[:s3_options]     || {}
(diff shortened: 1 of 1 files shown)


obahareth:
@jyurek I actually need this feature in a project of mine because in certain cases I switch my endpoints between Minio (as [mentioned in the docs](https://github.com/thoughtbot/paperclip/wiki/Paperclip-with-minio)) and S3. In some cases I need the same attachment to go to S3, and in another case I need it to go Minio. to Is there any other possible way to do this or should I make a new PR based on the solution this one provides (with tests included of course)?

obahareth:
I ended up using two different file objects and writing a method to get the relevant one based on my use-case.

Issue #1739, opened 2014-12-23
Just because Rails is defined does not mean Rails.env is


betesh:
Bug on https://github.com/thoughtbot/paperclip/blob/v4.2.1/lib/paperclip/storage/s3.rb#L291

See https://github.com/rails/rails-html-sanitizer/issues/25 and https://github.com/rails/rails-html-sanitizer/issues/26

jyurek:
Merged a fix for this. Thanks for the report!

PR #1742, opened 2015-01-05
Don't assume we have Rails.env if we have Rails

jyurek:
See #1739


lib/paperclip/rails_environment.rb
```diff
@@ -0,0 +1,23 @@
...
+      new.get
+    end
+
+    def get
+      if rails_exists? && rails_environment_exists?
+        Rails.env
+      else
+        nil
+      end
+    end
```
mike-burns: `private`

mike-burns:
I'm not opposed to making Paperclip more general and correct, though the idea of using Paperclip outside of Rails seems odd.

derekprior:
Can you add an explanation to the commit rather than a reference to an issue that itself is a reference to 2 other issues?

PR #1744, opened 2015-01-12
Use Travis container infrastructure

derekprior:
The container infrastructure (beta) has many advantages:
* More memory
* More CPU
* builds start much faster.
* builds run much faster (usually)

However, `sudo` cannot be used when running in this environment (which is
enabled with the `sudo: false` configuration.

The existing travis configuration had a `sudo` command that was added some time ago without much explanation. It's hoped that this is no longer necessary and that this PR will build okay without it...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ rvm:
 install:
   - "travis_retry bundle install"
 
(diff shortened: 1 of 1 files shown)


derekprior:
50% improvement!

jyurek:
:+1:

PR #1747, opened 2015-01-20
Added adapter to create Paperclip object from file path

nickjer:
Added an IO adapter that takes a string with the path to a file as opposed to an actual File object. It can be used as...

```
user.avatar = "images/bird.png"
```

We are using this in cases where a model or database is being populated from some user generated data files (i.e., yaml or csv). In these cases different models are used so knowing which variable is a file attachment is difficult to keep track of during the populating process.

Simplest solution would be setting each variable in the model to the corresponding string or value in the yaml file. That is where this adapter really shines.


jyurek:
I don't think this is a good adapter to be included in the gem. While it definitely seems useful for you, arbitrarily attempting to attach files based on a filename out of the box seems like it could easily lead to a security problem. Thank you for the PR, but I don't think this is a good fit for the gem.

jyurek:
You should package this as a separate gem, though, so those people who have needs like yours can still use it. It definitely seems useful.

nickjer:
No problem, thanks for the comment.

As for the security risk, it has the same vulnerabilities as the `http_url_proxy_adapter.rb` and `data_uri_adapter.rb` except instead of a URL path it is a file path. I could possibly do what Firefox does and use:

```
user.avatar =  'file:///home/username/www/images/bird.png'
```

PR #1755, opened 2015-01-28
Add `length` alias for `size` method in AbstractAdapter

dmitry:
The reason behind this I'm trying to use `StringioAdapter` inside the `ActionMailer`, and it fails because `length` used in one of the `mail` gem:
https://github.com/mikel/mail/blob/master/lib/mail/message.rb#L1997

Also you can see that normally `size` has `length` alias, for example in a `StringIO` class:
http://ruby-doc.org/stdlib-1.9.3/libdoc/stringio/rdoc/StringIO.html#method-i-length

Let me know, if this change requires test case, I will add it.


jyurek:
Yes, all code should be tested -- if for nothing else but regression catching. Thanks for adding this.

dmitry:
@jyurek did it. Hopefully now it's fine.

jyurek:
This is on master. I guess I didn't update this last week. Thanks!

PR #1802, opened 2015-03-06
Update Appraisals

maclover7:


maclover7:
@jferris @jyurek Ready to merge?

mike-burns:
I cannot get Appraisal to work. @jferris @sikachu what am I doing wrong?

```
% bundle exec appraisal install
>> bundle check --gemfile='/home/mike/paperclip/gemfiles/3.2.gemfile' || bundle install --gemfile='/home/mike/paperclip/gemfiles/3.2.gemfile'
Your Gemfile lists the gem paperclip (>= 0) more than once.
You should probably keep only one of them.
While it's not a problem now, it could cause errors if you change the version of just one of them later.
You cannot specify the same gem twice coming from different sources.
You specified that paperclip (>= 0) should come from source at ../../ and source
at ../
Your Gemfile lists the gem paperclip (>= 0) more than once.
You should probably keep only one of them.
While it's not a problem now, it could cause errors if you change the version of just one of them later.
You cannot specify the same gem twice coming from different sources.
You specified that paperclip (>= 0) should come from source at ../../ and source
at ../
```

maclover7:
@mike-burns That's weird - when Travis builds it is green...

adamniedzielski:
I am facing the very same issue as @mike-burns

jyurek:
This worked without issue for me, so I pulled it in. Thanks!

PR #1805, opened 2015-03-08
Introduce About thoughtbot section

cpytel:
* Move the thanks to contributors to the contributing section
* Introduce the new "About thoughtbot" section.


jferris:
Looks good.

Issue #1530, opened 2014-05-02
`file` detects wrong mime type for Open Office XML files


scpike:
Paperclip relies on the `file` command to detect the mime type of uploads, rather than the extension, since the extension is provided by the user. Unfortunately `file` reports "application/zip;" for pptx, docx, and xlsx files, so Paperclip thus stores the files with content_type = 'application/zip' rather than the correct mime types.  These files are technically zip files, but that's just an implementation detail that users don't care about - they want the files to open in the right programs. 

These seem like common enough file types that it's worth figuring out a solution. I'm happy to help with the code.

maclover7:
Hi everybody! Is this still an issue for you in Paperclip; I know this issue is from approximately 1 year ago. If it is still an issue, would you be willing to send in a PR to solve the problem? Thanks!

scpike:
@maclover7 this is still an issue for me, and I've been using my fork (https://github.com/scpike/paperclip/tree/ooxml-content-types) for a while. Looks like @jeremywadsack rebased that on top of the latest code and submitted a PR. That code might not pass CI, though, because I'm also using a mimemagic fork.

jeremywadsack:
@scpike I opened a PR from my fork because it was based on newer paperclip (but you are still marked as author). 

Thanks for the reminder about the mimemagic fork. I had forgotten that. I updated the PR to include that in the `Gemfile` which I hope fixes the CI errors in Travis. That said, I wonder is this is really the best solution or if we should use that fork in the `gemspec`.

scpike:
We should get the changes into mimemagic itself, probably, but I haven't submitted a PR there. Mimemagic is based off of the freedesktop.org file type detection, so I tried to get the openxml stuff merged at the source, but they haven't accepted my patch (for reasonable reasons): https://bugs.freedesktop.org/show_bug.cgi?id=78797.

jeremywadsack:
@scpike, yes, better to get the mimemagic changes there rather than in a fork. I'm using your mimemagic fork in production with no other changes, so it seems that would be the best route. It's simpler than the upstream patch which isn't going to fly anyway.

FWIW, though, adding your fork to the Gemfile on #1807  didn't help the Travis builds. Can you take a look at that and see if that's related to mimemagic or something else? I'm not really that familiar with this code.

maclover7:
@jferris @jyurek Please close issue, problem appears to be solved.

anykeyh:
[...]

PR #1807, opened 2015-03-10
Add support for ooxml content types through the mimemagic gem

jeremywadsack:
For issue #1530


maclover7:
Nice PR! Can you start out by fixing the Hound style guide issues?

jeremywadsack:
Mostly the code matches the current style. Although I can fix it if you're changing style. e.g.

```
 require 'mime/types'
+require 'mimemagic'
```

"Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping."

maclover7:
A) Not sure if we are going to be changing detectors, paging @jyurek @sikachu 
B) Some specs are failing on Travis, can you check that out?

scpike:
@maclover7 @jyurek @sikachu you'll want to check out the thread on #1530 for details, specifically https://github.com/thoughtbot/paperclip/issues/1530#issuecomment-43375497. I do think it's important that paperclip has a way to properly detect xlsx files, and I don't think it's good for paperclip to rely on specific newer versions of executables like `file`, but my fork works for my purposes so I'm not actively in any pain. 

@jeremywadsack The xlsx test is possibly failing in your PR because it gets it from the gemspec, not the Gemfile. I don't think you can use git sources in gemspecs.

maclover7:
@jeremywadsack @scpike I'm sorry, I don' think we can use personal forks of projects in this gem. Would you be willing to make a PR to mimemagic, and then we can use it?

scpike:
@maclover7 agreed, please don't actually use my fork. I'll make a PR against mimemagic in which I explain the situation. I'd avoided that because it means convincing them they should layer my magic on top of the magic they're getting from freedesktop.org. 

Is the team generally ok with the idea of using the magic tables from mimemagic?

scpike:
See https://github.com/minad/mimemagic/pull/19

jeremywadsack:
@scpike thanks for opening the mimemagic PR.

@maclover7 I totally agree. I shouldn't have considered a personal fork for a public project. :)

maclover7:
Is this ready to go again now - the minemagic PR was merged in

scpike:
@maclover7, @minad was kind enough to merge the support we need into his mimemagic library. A few things need to happen before you can merge this PR:

  - I need to update this code to use the new mimemagic functionality 

  - I need to figure out what's up with the tests / failing style

  - mimemagic needs to release a new gem version

jeremywadsack:
Stephen, if you need to update code, then I can close this PR and let you
open one.

Jeremy Wadsack

scpike:
@jeremywadsack sorry for the delay here. I'm going to find some time to look at this in the next couple of days. If I end up changing code I'll submit a new PR.

lib/paperclip/file_command_content_type_detector.rb
```diff
@@ -14,9 +14,9 @@ def detect
 
     def type_from_file_command
       type = begin
-        # On BSDs, `file` doesn't give a result code of 1 if the file doesn't exist.
-        Paperclip.run("file", "-b --mime :file", :file => @filename)
-      rescue Cocaine::CommandLineError => e
+        MimeMagic.by_magic(File.open(@filename)).try(:type) ||
+          Paperclip.run("file", "-b --mime :file", :file => @filename)
```
jyurek: Since this method is `type_by_file_command` it would make sense to extract this out into another method.

minad: Why not use mimemagic by default?

Am 24. April 2015 16:35:42 MESZ, schrieb Jon Yurek
<notifications@github.com>:

minad: Why not use mimemagic by default?

Am 24. April 2015 16:35:42 MESZ, schrieb Jon Yurek <notifications@github.com>:

jyurek: Well, this change makes it that way, since it comes first. `file` would be a fallback here.

jyurek:
@scpike Any movement here?

jyurek:
I merged in #1839 so this is not needed anymore. Thanks, though!

jeremywadsack:
Yay!

PR #1811, opened 2015-03-11
Fixes #1453

### LINKED ISSUE #1453: If call @model.file.url() on a non saved istance get a url instead of nil
Hi,

In the `action` view have
```
```

byterussian:


maclover7:
Hi @byterussian ! Can you edit your tests to make sure everything is passing? Thanks!

spec/paperclip/storage/s3_spec.rb
```diff
@@ -419,11 +429,13 @@ def original_filename
         production: { bucket: "prod_bucket" },
         development: { bucket: "dev_bucket" }
       },
+      bucket: "bucket",
```
byterussian: Anyone can help me with this? Maybe:
```
      rebuild_model storage: :s3,
                    s3_credentials: {
                      production: { bucket: "prod_bucket" },
                      development: { bucket: "dev_bucket" }},
                    bucket: "bucket",
                    s3_host_alias: "something.something.com",
                    path: ":attachment/:basename:dotextension",
                    url: ":s3_alias_url"
      @dummy = Dummy.new
      @dummy.avatar = stringy_file
      @dummy.stubs(:new_record?).returns(false)
```

byterussian:
Hi @maclover7, now all the tests pass.

jyurek:
@byterussian If you could squash these commits together, I'll gladly pull this in. Thanks!

PR #1813, opened 2015-03-12
remove useless condition

lancedikson:
It's useless because there is the same conditional in https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/railtie.rb#L25, and if ActiveRecord is not defined, Glue will not run.


maclover7:
Nice PR! Looks ready to merge to me.

@jferris @jyurek Ready to merge?

jyurek:
Merged in. Thanks!

PR #1817, opened 2015-03-13
Fixes #1453 (Squashed)

### LINKED ISSUE #1453: If call @model.file.url() on a non saved istance get a url instead of nil

byterussian:


spec/support/mock_model.rb
```diff
@@ -1,2 +1,5 @@
 class MockModel
```
jyurek: Why are you using MockModel when we already have a FakeModel, anyway?

byterussian: Seemed to me that both were used for the tests, I chose randomly. But it seems that MockModel it is used in [only one test](https://github.com/thoughtbot/paperclip/blob/master/spec/paperclip/url_generator_spec.rb#L46). Don't merge, I prepare two new pull request. One with the fixed code for issue #1453, another to remove completely `MockModel`, is useless(right?).

jyurek: Yeah, it appears useless, unless you hit resistance while removing it.

PR #1818, opened 2015-03-13
Fixes #1453 (Squashed)

### LINKED ISSUE #1453: If call @model.file.url() on a non saved istance get a url instead of nil

byterussian:


maclover7:
Hi @byterussian! Can you please fix the Hound style guide comments? Thanks!

byterussian:
Done, there is a method to squash the commits without open a new PR? I've already squashed and pushed with --force option but nothing.

maclover7:
@byterussian How did you try and squash?

byterussian:
Never mind, I've just tried again and now works. :smile:

maclover7:
@jferris @jyurek @mike-burns - Thoughts on this?

mike-burns:
Link to the bug: #1453.

Returning `nil` seems rather unpleasant. I'm tempted to say that the solution to the bug is "don't do that."

(And as the author of Wrapped, I liked to advocate for #url returning a wrapped object instead.)

@jferris @jyurek do you have more opinions on this?

byterussian:
Hi @mike-burns, in ActiveRecord a `Model.new` object returns `nil` for almost every attribute/field, IMHO there is no reason for paperclip to returns a nonexistent url.

jyurek:
While I agree with @mike-burns in principle, the reality is that we use `nil` all the time, and, IMHO, not using it here would be more surprising than using it. This looks ok to me, and once the tests are done running, I'll pull it in.

jyurek:
Merged this in. Thanks!

fw42:
I believe this PR might have introduced a new bug: https://github.com/thoughtbot/paperclip/issues/1930

PR #1821, opened 2015-03-22
Rescue only NotIdentifiedByImageMagickError in Attachment#post_process_style

### LINKED ISSUE #1352: CommandNotFoundError is rescued and added to model errors
I spent some time debugging when I pushed the code to my CI server and one spec was failing. The spec was a Capybara feature which was submitting form with attachment and checking if the model got saved. Finally I found the error among validation errors - it was CommandNotFoundError - "Could not run the `identify` command. Please install ImageMagick."

This was quite surprising, because it is a missing dependency and certainly not validation error. I cannot imagine a case when I want to present it to the end-user. 

Looking through the code I found the reason of this behaviour:

https://github.com/thoughtbot/paperclip/blob/adddf87cf30d416d71c7b2a80a004d170d10dc2c/lib/paperclip/attachment.rb#L434-437

In this case I would expect that the exception is not rescued at all because it is a matter of overlooked dependency and should be clearly visible. Visible to the dev, not the user. 

What do you think?

adamniedzielski:
Other errors should bubble up, because they are meaningless for end-user, but important for programmer.

Fixes #1352

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -528,7 +528,7 @@ module Paperclip
         @queued_for_write[name] = Paperclip.io_adapters.for(@queued_for_write[name])
         unadapted_file.close if unadapted_file.respond_to?(:close)
         @queued_for_write[name]
-      rescue Paperclip::Error => e
+      rescue Paperclip::Errors::NotIdentifiedByImageMagickError => e
         log("An error was received while processing: #{e.inspect}")
         (@errors[:processing] ||= []) << e.message if @options[:whiny]
       ensure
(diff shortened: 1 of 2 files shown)


maclover7:
Nice PR! Can you fix the Hound CI style guide comments? Thanks!

adamniedzielski:
@maclover7 Done :)

maclover7:
@jferris @jyurek Comments on this?

mike-burns:
Can you add a test for #1352, please?

maclover7:
@mike-burns There was a test included in the PR...

Ready to merge?

mike-burns:
The test isn't for #1352. This code change should include a new test, not (just) a modified existing test.

adamniedzielski:
@maclover7 @mike-burns I added a new test case.

mike-burns:
Thanks, merged as e195ef66978636ac93cab949bf322568ed291f1c.

PR #1833, opened 2015-04-14
Fix for Errno::ENAMETOOLONG in Thumbnails

codeodor:
* also includes a test for TempfileFactory to be sure it does not revert to similar behavior

### DIFF
--- a/lib/paperclip/thumbnail.rb
+++ b/lib/paperclip/thumbnail.rb
@@ -63,8 +63,8 @@ module Paperclip
     # that contains the new image.
     def make
       src = @file
-      dst = Tempfile.new([@basename, @format ? ".#{@format}" : ''])
-      dst.binmode
+      filename = [@basename, @format ? ".#{@format}" : ""].join
+      dst = TempfileFactory.new.generate(filename)
 
       begin
         parameters = []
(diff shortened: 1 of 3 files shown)


codeodor:
Fixed hound issues and rebased.

codeodor:
What do you think of using Paperclip's Tempfile here instead of Ruby's?

codeodor:
If this looks acceptable, I'll rebase it into a single commit. Let me know. Thanks!

jyurek:
Get your head in the game! I've seen a video of a guy soloing the whole raid in 12 minutes!

;)

Sammy Larbi wrote:

lib/paperclip/thumbnail.rb
```diff
@@ -63,8 +63,8 @@ def convert_options?
     # that contains the new image.
     def make
       src = @file
-      dst = Tempfile.new([@basename, @format ? ".#{@format}" : ''])
-      dst.binmode
+      filename = [@basename, @format ? ".#{@format}" : ""].join
+      dst = TempfileFactory.new.generate(filename)
```
jyurek: Do we not still need `binmode`?

jyurek:
Oh, duh, `TempfileFactory` handles that. Yeah, this looks good to me, if you could rebase, that would be awesome.

lib/paperclip/thumbnail.rb
jyurek: Oh, duh. `TempfileFactory` handles that.

codeodor:
Ok, done!

jyurek:
Thanks!

codeodor:
My pleasure :+1: and thanks for Paperclip! :trophy:

Issue #1729, opened 2014-12-16
(4.2.1) `Paperclip::Storage::S3#copy_to_local_file` has size 0


slepkin:
This bug appears in 4.2.1, but not 4.2.0.

Context: Our `Asset` object has a `pdf` attachment.

#### 4.2.1
```
2.1.5 :002 > asset = Asset.find(730203)
 => #<Asset id: 730203, created_at: "2014-12-16 19:50:27", updated_at: "2014-12-16 19:50:27", pdf_file_name: "pdf_e0ab3c21-310f-4ee8-811c-6165438ff9e1.pdf", pdf_content_type: "application/pdf", pdf_file_size: 61458, pdf_updated_at: "2014-12-16 19:50:27"> 
2.1.5 :003 > asset.pdf.size
 => 61458 
2.1.5 :004 > asset.pdf.copy_to_local_file(:original, "tmp/test.pdf")
[paperclip] copying assets/pdfs/000/730/203/original/pdf_e0ab3c21-310f-4ee8-811c-6165438ff9e1.pdf to local file tmp/test.pdf
[AWS S3 200 0.777251 0 retries] get_object(:bucket_name=>"assets",:key=>"assets/pdfs/000/730/203/original/pdf_e0ab3c21-310f-4ee8-811c-6165438ff9e1.pdf")  

 => "%PDF-1.4\n1 0 obj\n..." # Long string
2.1.5 :005 > File.open("tmp/test.pdf").size
 => 0 
```

#### 4.2.0
```
2.1.5 :002 > asset = Asset.find(730203)
 => #<Asset id: 730203, created_at: "2014-12-16 19:50:27", updated_at: "2014-12-16 19:50:27", pdf_file_name: "pdf_e0ab3c21-310f-4ee8-811c-6165438ff9e1.pdf", pdf_content_type: "application/pdf", pdf_file_size: 61458, pdf_updated_at: "2014-12-16 19:50:27"> 
2.1.5 :003 > asset.pdf.size
 => 61458 
2.1.5 :004 > asset.pdf.copy_to_local_file(:original, "tmp/test.pdf")
[paperclip] copying assets/pdfs/000/730/203/original/pdf_e0ab3c21-310f-4ee8-811c-6165438ff9e1.pdf to local file tmp/test.pdf
[...]

PR #1730, opened 2014-12-17
Aws sdk v2

danielwanja:
* Added s3 storage implementation using the aws-sdk v2.

PR #1838, opened 2015-04-23
Use Updated Version of `aws-sdk` in Docs [ci skip]

reshleman:
The `aws-sdk` development dependency was updated to `~> 1.6` in #1816
to resolve issues with the creation of 0-byte files on S3 (see #1729).
However, there is no runtime dependency on `aws-sdk`, and the
documentation still suggests that users install an earlier version of
`aws-sdk` that produces the problems in the referenced issue.

This commit updates the docs to recommend the use of `aws-sdk` `~> 1.6`
with Paperclip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -531,7 +531,7 @@ You may also choose to store your files using Amazon's S3 service. To do so, inc
 the `aws-sdk` gem in your Gemfile:
 
 ```ruby
-gem 'aws-sdk', '~> 1.5.7'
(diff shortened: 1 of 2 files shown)


waleedasif322:
Any plan to update to `aws-sdk` v2? If not, I can try to do a PR

reshleman:
@waleedasif322 Looks like there's an open PR for for `aws-sdk` version 2.x: #1730.

PR #1839, opened 2015-04-26
Prefer the mimemagic gem for content type detection

scpike:
Prioritize MimeMagic over the `file` binary for content type detection
when it finds a match. Fall back to `file` if MimeMagic can't match
anything.

`file` incorrectly detects Open Office XML files (e.g., xlsx, docx) as
zip since they're implemented as zipped archives of xml files. MimeMagic
detects them properly.

This should be in place of #1807


scpike:
@jyurek please check this out in place of #1807. #1807 was @jeremywadsack's fork of my original code, which was never meant to be pushed to the core project. I've fixed some style issues, added the released `mimemagic` gem with support for OOXML files, and moved the MimeMagic detection outside of the FileCommandContentTypeDetector class. I didn't make a new class for it since it's just one line of code, so let me know if you want me to re-arrange things in some way that aligns with the project's architecture better.

I had to stub some stuff out wherever the tests test non-existent files by stubbing out `Paperclip.run`. 

The remaining hound violations don't seem consistent with the project so I didn't address them.

jyurek:
Merged. Thanks a lot!

aharbick:
it seems like this trick needs to be replicated in paperclip/media_type_spoof_detector.rb

or am I missing something?

jeremywadsack:
That's a good question. I don't actually see any reason that the `paperclip/media_type_spoof_detector.rb` isn't taking advantage of the `content_type_detector`. Maybe @jyurek understands better.

etagwerker:
@jeremywadsack +1 only if that also means that we will use `file` as fall back. 

I just found a related issue when trying to validate the mime type for a Javascript file attachment. [see: https://github.com/minad/mimemagic/issues/30]

scpike:
It was not an intentional decision on my part – I didn't think to check the rest of the project for other uses for the file command when I made this PR. :+1: for using my logic (which has the file fallback) everwhere you want to do content type detection.

PR #1853, opened 2015-05-12
allow sublcasses to override definitions

midu:
It seems like when redefining attachment definitions in STI classes, only the styles from the parent class are being picked up. 

```rb
class Photo < ActiveRecord::Base
  has_attached_file :data, styles: { small: "70x70#" }
# ...


class UserPhoto < Photo
  has_attachment_file :data, styles: { small: "100x100>" }
# ...
```

Expectation:

```rb
UserPhoto.attachment_definitions[:data][:styles] == { small: "100x100>" }
```

Reality: 

```rb
UserPhoto.attachment_definitions[:data][:styles] == { small:  "70x70#" }
```

That seems like a strange behavior, and it definitely didn't do that in earlier versions of paperclip (like 3.4.0, where we're upgrading from).

This PR addresses this issue, let us know what you think.

cc-ing @mgerrior who found and fixed the issue

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -7,6 +7,7 @@ require 'active_support/core_ext'
 require 'mocha/api'
 require 'bourne'
 require 'ostruct'
+require 'pathname'
 
 ROOT = Pathname(File.expand_path(File.join(File.dirname(__FILE__), '..')))
 
--- a/lib/paperclip/attachment_registry.rb
+++ b/lib/paperclip/attachment_registry.rb
@@ -51,7 +51,7 @@ module Paperclip
     end
 
     def definitions_for(klass)
-      klass.ancestors.each_with_object({}) do |ancestor, inherited_definitions|
+      klass.ancestors.reverse.each_with_object({}) do |ancestor, inherited_definitions|
         inherited_definitions.deep_merge! @attachments[ancestor]
       end
     end
--- a/lib/paperclip/attachment_registry.rb
+++ b/lib/paperclip/attachment_registry.rb
@@ -51,7 +51,8 @@ module Paperclip
     end
 
     def definitions_for(klass)
-      klass.ancestors.reverse.each_with_object({}) do |ancestor, inherited_definitions|
+      parent_classes = klass.ancestors.reverse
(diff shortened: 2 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -7,6 +7,7 @@
 require 'mocha/api'
 require 'bourne'
 require 'ostruct'
+require 'pathname'
```
midu: This seems to be required in ruby 2.2+

midu:
I see all these comments from hound, happy to fix them but I really was following the styles of the files I were in. Let me know what you prefer

midu:
Since we're on the subject, I am wondering how much you think the `deep_merge` in that method makes sense...

If I start overriding my styles in my subclass, I do not think that most people expect to have some styles that they did not explicitly defined in the subclass.

```rb
class Photo < ActiveRecord::Base
  has_attached_file :data, styles: { small: "70x70#", medium: "500x500" }
# ...


class UserPhoto < Photo
  has_attachment_file :data, styles: { large_avatar: "100x100>", small_avatar: "35x35" }
# ...
```

To me that's a typical use of STI for images where you use an image class as a polymorphic association and specialize some versions of the image to some models specific needs. 

With the current implementation, UserPhoto will actually create 4 styles, small, medium, large_avatar, small_avatar.

While I can kind of understand the `merge` for the attachment options (things like `url`, `whiny`, `keep_old_styles` are most likely going to be the same for the subclasses), the merge of `styles`, or `convert_options` to make a lot less sense, and I would guess that use STI with variations of style don't even realize what it actually does.

midu:
Anybody interested in this PR? I believe it's pretty important.

I am happy to answer any questions about this if anything isn't clear

tute:
@midu hopefully tomorrow Friday @jyurek helps me doing some PR reviews. Sorry we've been busy and couldn't respond timely.

midu:
@tute no worries, didn't mean to rush you into this, just wanted to make sure someone saw it :)

thanks for replying!

tute:
:+1:

jyurek:
Merged in. Thanks!

PR #1858, opened 2015-05-15
Add single and double quotes to restricted characters for an attachment

charlesdg:
Single and doubles quotes in the attachment file name can lead to issues when the url of the attachment is in a single or double quotes string.

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -14,7 +14,7 @@ module Paperclip
         :default_style         => :original,
         :default_url           => "/:attachment/:style/missing.png",
         :escape_url            => true,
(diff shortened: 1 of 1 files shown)


jyurek:
Issues like what?

Also, can you add/modify tests for this? Thanks.

charlesdg:
like when you use it in an url in a CSS rule: 
`background-image: url('http://mydomain.com/namewithquote's.jpg');`

I will check for the test !

jyurek:
That makes a lot of sense. Would love to see the tests for this so it gets in.

tute:
File names with quotes are valid UNIX file names. I don't think they should be disallowed. Thanks for your input and sharing your work! ❤️

PR #1859, opened 2015-05-15
Use scheme for fog expiring url

f1sherman:
Honor the configured scheme for fog expiring URLs. In order to be consistent with public URLs, this also changes the default behavior for expiring URLs to use the https scheme rather than http.

Also included are a couple of minor changes due to issues I noticed trying to get the test suite running.


jyurek:
Merged in. Thanks!

Issue #1854, opened 2015-05-12
Documentation


maclover7:
Should we move documentation out of the README and into the wiki for better organization? Or should we just add a TOC -- I feel it is starting to get a big long and unorganized.

tute:
A toc in the README is best. wikis can go out of control, and people don't use the diff view as much as in PR reviews.

pandoc is a good tool to auto-generate TOCs on markdown files, there's a snippet somewhere in StackOverflow.

maclover7:
Sounds good, will take care of this and add a TOC over the weekend.

tute:
ping @maclover7! If you'd like we can take care of this too on Friday.

tute:
Some help (generated with vim plugin for markdown `Toc`):

```
Paperclip
  Requirements
    Ruby and Rails
    Image Processor
    `file`
  Installation
  Quick Start
    Models
    Migrations
    Edit and New Views
    Edit and New Views with Simple Form
    Controller
Use strong_parameters for attribute whitelisting
Be sure to update your create() and update() controller methods.
    Show View
    Deleting an Attachment
  ```
  Usage
  Validations
Security Validations
Allow ".foo" as an extension for files with the mime type "text/plain".
  Defaults
  ```
  Migrations
    Table Definition
    Schema Definition
    Vintage syntax
  Storage
    Understanding Storage
  Post Processing
  Events
  URI Obfuscation
  MD5 Checksum / Fingerprint
  File Preservation for Soft-Delete
  Custom Attachment Processors
  Dynamic Configuration
    Dynamic Styles:
    Dynamic Processors:
  Logging
  Deployment
  ```yml
  Testing
  ```
  Contributing
  License
  About thoughtbot
```

tute:
It wasn't pandoc, it's doctoc: https://github.com/thlorenz/doctoc

tute:
Closing in favor of PR. Thank you @maclover7!

PR #1870, opened 2015-05-25
Documentation TOC

maclover7:
Companion PR for #1854.


maclover7:
@tute Is this ready to go?

tute:
That is great! Go ahead, rebase `master` onto `maclover7-toc` (we don't use the merge button).

Did you do it by hand? Ideally it would be almost-automatically generated, so when sections changes calculating the diff is easy. I noticed that with current markup sections get skewed. It's not really important, just a nice to have.

Anyhow, rebase away! :+1:

maclover7:
Yeah, I did it by hand. Why do we not merge PRs?

tute:
I think this is mostly a style decision, and it might have some interesting technicalities while trying to revert commits (versus reverting trees merges). A rationale can be found in https://robots.thoughtbot.com/rebase-like-a-boss.

tute:
Thank you!

PR #1873, opened 2015-05-26
Decrease Memory footprint

schneems:
Mime-types 2.6.1+ has a columnar store that only loads the data that is needed for a drastically reduced memory footprint. cc/ @jeremyevans @halostatue


tute:
Related PR: https://github.com/mime-types/ruby-mime-types/pull/96.

Thank you, @schneems!

PR #1875, opened 2015-05-27
Release v4.3.0

maclover7:
PR to release v4.3.0 of Paperclip.

@tute and @jyurek, are we set to go here? All that's left to do is to merge PR, create & push git tag, and push up to rubygems.org


NEWS
```diff
@@ -1,3 +1,15 @@
+New in 4.3.0:
```
tute: Tiny one, let's add a new line here please.

paperclip.gemspec
```diff
@@ -18,10 +18,10 @@ Gem::Specification.new do |s|
   s.require_paths = ["lib"]
 
   s.requirements << "ImageMagick"
-  s.required_ruby_version = ">= 1.9.2"
+  s.required_ruby_version = ">= 2.0.0"
 
-  s.add_dependency('activemodel', '>= 3.0.0')
-  s.add_dependency('activesupport', '>= 3.0.0')
+  s.add_dependency('activemodel', '>= 3.2', '>= 4.1.0')
+  s.add_dependency('activesupport', '>= 3.2', '>= 4.1.0')
```
tute: I wonder how this work in the gemspec. We want a subset of two mayor version, we want two "equations". Would 4.0 comply with this specification? Something to research.

tute: There we go, we'll need some adjustments of our restrictions:

```
  In Gemfile:

    paperclip (>= 0) ruby depends on

      activesupport (>= 3.2, >= 4.1.0) ruby

    rails (~> 3.2.15) ruby depends on

      activesupport (3.2.15)
```

tute: I might have written it wrong, it seems like `>=` should be `~>`.

tute:
This looks good to me, will let @jyurek sign off.

tute:
We tend to do PRs as small as possible. In this case, I would have submitted two: the dependencies versions change, and the 4.3 release parts. This is small enough that it doesn't really matter anyway, but good to keep in mind! Thanks for your work.

maclover7:
@tute Try to re-run the 3.2 Travis tests (I can't for some reason) -- seems like the error may be on Bundler's end:

```
Retrying dependency api due to error (2/3): Bundler::HTTPError Net::HTTPUnprocessableEntity: Too many gems (use --full-index instead)
```

maclover7:
@tute Just squashed everything down into one commit, and I think tests should be passing now -- will wait to hear from Travis.

maclover7:
Build is green. Are we all set to go here? @tute @jyurek

tute:
This LGTM, hopefully it gets merged next Friday.

.travis.yml
```diff
@@ -1,5 +1,4 @@
 rvm:
-  - 1.9.3
```
jyurek: Why are we dropping support for 1.9.3? People still use that, and we should probably make sure Paperclip works on it, shouldn't we?

NEWS
```diff
@@ -1,3 +1,16 @@
...
+* Improvement: Update aws-sdk and cucumber gem versions.
+* Improvement: Add `length` alias for `size` method in AbstractAdapter.
+* Improvement: Removed some cruft
+* Improvement: deep_merge! Attachment definitions
+* Improvement: Switch to minemagic gem for content-type detection
+* Improvement: Allows multiple content types for spoof detector
+* Bug Fix: Don't assume we have Rails.env if we have Rails
+* Performance: Decrease Memory footprint
+* Ruby Versioning: Drop support for 1.9.3 (EOL'ed)
+* Rails Versioning: Drop support for 4.0.0 (EOL'ed)
```
jyurek: There seems to be a lot more in the git log then just this between 4.2.1 and here.

gemfiles/3.2.gemfile
```diff
@@ -8,7 +8,7 @@ gem "activerecord-jdbcsqlite3-adapter", :platforms => :jruby
 gem "rubysl", :platforms => :rbx
 gem "racc", :platforms => :rbx
 gem "pry"
-gem "rails", "~> 3.2.15"
+gem "rails", ">= 3.2.0"
```
jyurek: Why backslide on the Rails version here?

paperclip.gemspec
```diff
@@ -18,15 +18,15 @@ Gem::Specification.new do |s|
   s.require_paths = ["lib"]
 
   s.requirements << "ImageMagick"
-  s.required_ruby_version = ">= 1.9.2"
+  s.required_ruby_version = ">= 2.0.0"
 
-  s.add_dependency('activemodel', '>= 3.0.0')
-  s.add_dependency('activesupport', '>= 3.0.0')
+  s.add_dependency('activemodel', '>= 3.2.0', '>= 4.1.0')
```
jyurek: Seems these should be `"~>3.2.0", "~>4.1"` right? Otherwise `">= 3.2.0"` will still cover Rails 4.0.

jyurek:
I had a few questions, but if nothing else I'll be able to push this out Monday. Thanks for pushing me on this.

.travis.yml
tute: > Why are we dropping support for 1.9.3? People still use that, and we should probably make sure Paperclip works on it, shouldn't we?

1.9.3 reached its end of life, and doesn't receive any patches any longer. Should we depend on software that doesn't receive security fixes any more?

NEWS
tute: > There seems to be a lot more in the git log then just this between 4.2.1 and here.

If there's relevant changes we should add them, otherwise I'm find with a higher signal-to-noise (and less data) in the NEWS. I'd use a CHANGELOG file (or the git log) to see the full detail.

gemfiles/3.2.gemfile
tute: > Why backslide on the Rails version here?

No real reason. Bundler will always install the highest patch version anyway.

paperclip.gemspec
tute: > Seems these should be "~>3.2.0", "~>4.1" right? Otherwise ">= 3.2.0" will still cover Rails 4.0.

Pretty sure that's the case.

tute:
Thanks for your review and work Jon!

.travis.yml
jyurek: It's not that we "depend" on it, it's that we support it.

tute: Indeed. Corrected question still stands though: should we support EOL'd software? Users can always continue to use Rails 4, and Ruby 1.9, and paperclip <= 4.2.1.

paperclip.gemspec
jyurek: Just tried this out, and it doesn't work. The conditions are AND, and you can't install something that satisfies both here. I think we're going to have to stick with `>= 3.2.0`.

NEWS
```diff
@@ -1,3 +1,16 @@
+New in 4.3.0:
+
+* Improvement: Update aws-sdk and cucumber gem versions.
+* Improvement: Add `length` alias for `size` method in AbstractAdapter.
+* Improvement: Removed some cruft
+* Improvement: deep_merge! Attachment definitions
+* Improvement: Switch to minemagic gem for content-type detection
```
jyurek: `minemagic` ;)

jyurek:
@maclover7 Few more comments. I'm waiting on an external issue to be resolved, but this looks good otherwise.

.travis.yml
jyurek: Yeah I guess that's fair. They already know they're running on thin ice by using that Ruby and/or Rails.

maclover7:
Are we ready to go here? @jyurek @tute

jyurek:
There's a few things I found: A typo, and the gem versions won't work as-is. But the concept of releasing 4.3.0 is solid.

tute:
@maclover7 please address the PR comments, hopefully we release today.

tute:
Status update: @maclover7 we released 4.2.2 with a security fix, and we'll hold off for a few days on 4.3 because master has a red CI, and Jon can't work on it today.

lastobelus:
with this commit, can not bundle update a rails 4.1 app:

```
#Gemfile
gem 'rails', '~> 4.1.10'
gem 'paperclip', github: 'thoughtbot/paperclip'
```
```
Bundler could not find compatible versions for gem "activemodel":
  In Gemfile:
    paperclip (>= 0) ruby depends on
      activemodel (~> 3.2.0) ruby

    rails (~> 4.1.10) ruby depends on
      activemodel (4.1.10)
```

tute:
That's surprising, I can't see how is paperclip filtering out activemodel 4.1.x. Can you spot it?

lastobelus:
I think in paperclip.gemspec that line 23 needs to be similar to line 29.

tute:
Yep! Seemingly master is not declaring dependencies right, I don't understand why as this PR fixes that?

https://github.com/thoughtbot/paperclip/blob/ecdd5427437f1ea0061e611b459d48a4adaf2ca7/paperclip.gemspec#L23-L29

Thanks for your info @lastobelus. @jyurek it seems we need to allow for >3.2, not ~>.

tute:
> with this commit, can not bundle update a rails 4.1 app:

Just released paperclip 4.2.3, should work well now!

tute:
Released 4.3.0, master is released at last! https://rubygems.org/gems/paperclip

maclover7:

PR #1877, opened 2015-05-27
Check filesystem for existence, instead of just instance db column. Closes #1876

### LINKED ISSUE #1876: .present? relies on a file_name being present. Not enough
I ran into a bug in our app where we were updating the filename column, but accidentally deleting the underlying file (by assigning the paperclip property to nil). `Paperclip::Attachment#present?` still returns true even though the file doesn't exist. 

I propose changing the method to use `File.exists?` as well.

mehulkar:


tute:
Copying from related issue https://github.com/thoughtbot/paperclip/issues/1876:

> I ran into a bug in our app where we were updating the filename column, but accidentally deleting the underlying file (by assigning the paperclip property to nil). `Paperclip::Attachment#present?` still returns true even though the file doesn't exist. 
>
> I propose changing the method to use `File.exists?` as well. 

---

Can you please add a regression test? Thank you!

mehulkar:
Sure! Wanted to put up a proof of concept for feedback before putting in more time.

mehulkar:
This PR breaks a lot of tests and I'm not sure why. 

I investigated a little bit it and it seems like it's the validation that's failing. Specifically the media type spoof validation that is looking to run the `file` command, but since the `AbstractAdapter` is being used (possibly a stub for the test suite?) the file that it attempts to run on is actually nil. So `nil.path` fails. 

I don't understand why/where this PR is causing this though. 

@tute could you point me in the right direction?

tute:
@mehulkar I've been helping with the issue tracker and some code reviews but actually don't yet know of paperclip internals. Will delegate to @jyurek.

jyurek:
I'm sorry, this isn't going to work, at least in its current form. You modified the method in `attachment.rb` which means it would be global to all Paperclip attachments. Files being on the local filesystem is (as far as we use it) the least common way of using Paperclip. This would flat out fail in all cases for S3 attachments (or any other storage system anyone uses).

If you want to check the storage backend for the file's actual existence, please use `exists?`.

Issue #1696, opened 2014-11-12
Validations fail for seemingly valid images


acconrad:
Upgrading from 2.7 to 4.2 on Rails 3.2.20 (Ruby 2.1.2).

In our model...

```
validates_attachment :profile_photo,
    content_type: { :content_type => ["image/jpeg", "image/jpg", "image/png", "image/gif"] },
    file_name: { :matches => [/png\Z/, /jpe?g\Z/, /gif\Z/] }
```

In our `spec/factories.rb` file...

```
include ActionDispatch::TestProcess
FactoryGirl.define do

  trait :with_profile_photo do
    profile_photo { fixture_file_upload(Rails.root.join('spec', 'support', 'test_image.jpg'), 'image/jpeg') }
  end
end
```

but when creating a factory, such as `create(:user, :with_profile_photo)` we're getting the following error:

> ActiveRecord::RecordInvalid:
>        Validation failed: Profile photo has an extension that does not match its contents

I have verified that our `test_image.jpg` does indeed have the `image/jpeg` content-type and is a valid JPG. Not sure what else we would need to do to get this to pass validations, the validations seem in line with what is being tested.

AteqEjaz:
Hi,

Facing exactly the same issue,

Validation failed: Profile pic has an extension that does not match its contents

Thanks!

tute:
@acconrad is this still an issue for you? How can we replicate it?

acconrad:
@tute this is still an issue for us. I think one mistake I made was that I reported us using `2.7` of Paperclip but we are actually running over version 2.6. 

So in order to replicate you would need to currently have Paperclip 2.6 running with Rails 3.2.21 and Ruby 2.1.5.

We currently have the following set up in a model:

```ruby
  has_attached_file :profile_photo,
    styles: {
      small: ['100x100>', :jpg],
      large: ['300x300>', :jpg],
      profile: ['300x300#', :jpg],
      search: ['90x90#', :jpg],
      very_large: ['500x500>', :jpg]
    },
    hash_secret: '9rleB3FN2qAkjnH0uVhsKozJg1PacDO',
    url: '/profile/:hash.:extension',
    storage: :s3,
    s3_credentials: "#{Rails.root}/config/s3.yml",
    s3_protocol: 'https',
    default_url: 'missing-profile-photo.png'

validates_image :profile_photo
```

and in our `spec/factories.rb` file for the model we have a trait:

```ruby
trait :with_profile_photo do
[...]

PR #1892, opened 2015-06-09
V4.2.2 fix spoof detector

fxtentacle:
Instead of assuming that the file command will always return exactly one mime type, this can handle 2 results if file returns them. It will still break when file finds 3+ mime type matches because file will not return the 3rd or following mime types (due to a big in libmagic). Still, this should greatly reduce the likelyhood for issue #1696


jyurek:
This looks reasonable to pull in. Can you add tests and take care of the Hound issues?

tute:
@fxtentacle: Can you fix the tests and address Hound comments? Thanks for your work! :heart:

PR #1895, opened 2015-06-10
If Rails is defined but Rails::Railtie is not, don't try to load the Railtie

betesh:
See https://github.com/rails/rails/pull/18196 and https://github.com/rails/rails-html-sanitizer/issues/26 for a discussion of why this is necessary.

Not sure why this is failing on travis but it doesn't look like it's because of this change.

### DIFF
--- a/lib/paperclip.rb
+++ b/lib/paperclip.rb
@@ -69,7 +69,7 @@ require 'mimemagic/overlay'
 require 'logger'
 require 'cocaine'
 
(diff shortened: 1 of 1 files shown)


betesh:
rebased onto the latest master

tute:
I'm ok upon green CI on this. Will let @jyurek take the lead anyway.

jyurek:
Thanks, @betesh!

PR #1897, opened 2015-06-12
Fixes

jyurek:
Fixes the gemspec dependencies.
Fixes failing specs.
`bundle exec appriasal rake` is green for me after this.

### DIFF
--- a/gemfiles/3.2.gemfile
+++ b/gemfiles/3.2.gemfile
@@ -2,13 +2,13 @@
 
 source "https://rubygems.org"
 
-gem "sqlite3", "1.3.8", :platforms => :ruby
+gem "sqlite3", "~> 1.3.8", :platforms => :ruby
 gem "jruby-openssl", :platforms => :jruby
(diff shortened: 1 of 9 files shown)


gemfiles/3.2.gemfile
```diff
@@ -8,7 +8,7 @@ gem "activerecord-jdbcsqlite3-adapter", :platforms => :jruby
 gem "rubysl", :platforms => :rbx
 gem "racc", :platforms => :rbx
 gem "pry"
-gem "rails", ">= 3.2.0"
+gem "rails", "~> 3.2.15"
```
tute: Oh right, this needs to be restricted. I typically leave the zero so we don't go crazy with "is this the latest patch release?", and we make it explicit that we expect bundler to install latest.

tute:
I'd open this PR agains `v4.2` branch, and then merge master onto that. These changes are needed in both places.

tute:
Versions look good to me. On green CI I'd merge in that order.

Issue #1764, opened 2015-02-09
AWS Version 2 support - uninitialized constant Paperclip::Storage::S3::AWS


fuzzyjulz:
It appears that the latest version of AWS - version 2, is not compatible with paperclip. From the aws-sdk-ruby:

> NameError: uninitialized constant AWS
>
>If you receive this error, you likely have upgraded to version 2 of the aws-sdk gem unintentionally. Version 2 uses the Aws namespace, not AWS. This allows version 1 and version 2 to be used in the same application.
> http://ruby.awsblog.com/post/TxFKSK2QJE6RPZ/Upcoming-Stable-Release-of-AWS-SDK-for-Ruby-Version-2

Is there a plan to do the upgrade?

darylprickett:
In the meantime the readme should probably be changed from 
gem 'aws-sdk', '~> 1.5.7'
to
gem 'aws-sdk', '< 2.0'

More information:
http://ruby.awsblog.com/post/TxFKSK2QJE6RPZ/Upcoming-Stable-Release-of-AWS-SDK-for-Ruby-Version-2

hughevans:
How to switch to fog:

```diff
   config.upload_file_options = {
-    storage: :s3,
-    s3_credentials: {
-      bucket:            ENV['FOG_DIRECTORY'],
-      access_key_id:     ENV['AWS_ACCESS_KEY_ID'],
-      secret_access_key: ENV['AWS_SECRET_ACCESS_KEY']
+    storage: :fog,
+    fog_credentials: {
+      aws_access_key_id: ENV['AWS_ACCESS_KEY_ID'],
+      aws_secret_access_key: ENV['AWS_SECRET_ACCESS_KEY'],
+      provider: ENV['FOG_PROVIDER'],
+      region: ENV['FOG_REGION'],
+      scheme: 'https'
     },
-    s3_headers: {
+    fog_directory: ENV['FOG_DIRECTORY'],
+    fog_file: {
       'Cache-Control' => 'max-age=315576000',
       'Expires'       => 10.years.from_now.httpdate
     },
-    url: ':s3_alias_url',
-    s3_host_alias: '/',
     path: 'files/:hash_:style.:extension',
     hash_secret: ENV['SECRET_KEY_BASE']
   }
```

fuzzyjulz:
Hmmm but that is just changing the implementation for the Fog one.

[...]

PR #1903, opened 2015-06-17
The long awaited AWS v2 support PR

betesh:
Thank you to @davetchen, who really did most of the work here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ gemfile:
   - gemfiles/3.2.gemfile
   - gemfiles/4.1.gemfile
   - gemfiles/4.2.gemfile
+  - gemfiles/3.2.awsv1.gemfile
+  - gemfiles/4.1.awsv1.gemfile
+  - gemfiles/4.2.awsv1.gemfile
 
 matrix:
   fast_finish: true
--- a/Appraisals
+++ b/Appraisals
@@ -1,11 +1,29 @@
 appraise "3.2" do
   gem "rails", "~> 3.2.0"
+  gem "aws-sdk", "~> 2.0"
 end
 
 appraise "4.1" do
   gem "rails", "~> 4.1.0"
+  gem "aws-sdk", "~> 2.0"
 end
 
 appraise "4.2" do
   gem "rails", "~> 4.2.0"
+  gem "aws-sdk", "~> 2.0"
+end
+
+appraise "3.2.awsv1" do
+  gem "rails", "~> 3.2.0"
+  gem "aws-sdk", "~> 1.5"
+end
+
+appraise "4.1.awsv1" do
+  gem "rails", "~> 4.1.0"
+  gem "aws-sdk", "~> 1.5"
+end
+
+appraise "4.2.awsv1" do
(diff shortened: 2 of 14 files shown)


tute:
I was about to release v4.3, but this should part of the release, right @jyurek?

tute:
I'd love if we could use an adapter instead of adding conditionals everywhere we use AWS in the code.

tute:
There's too many releases in 4.3, will release anyway, and we can add this in 4.4.

betesh:
> I'd love if we could use an adapter instead of adding conditionals everywhere we use AWS in the code.

@tute, @trevorrowe proposed something similar (https://github.com/thoughtbot/paperclip/issues/1764#issuecomment-103325634) but proposed leaving the conditions in place (https://github.com/thoughtbot/paperclip/issues/1764#issuecomment-103569165) since AWS v1 will soon be deprecated anyway.

betesh:
@tute  FYI I have added more Gemfiles so that travis tests both AWS versions.  Everything is passing on both versions.

This makes it very safe to add additional features to this branch that are specific to AWS v2, as long those features are accompanied by appropriate tests.  Would any of the #1730 contributors like to try merging the new functionality there into this PR?

tute:
cc @jyurek to merge this PR

jyurek:
I had some local failures related to regions, so I put in a fix for them, but this is merged. There are things I would clean up in the code, but I can't let that stop this PR at this point. We can fix them in the future. Thank you very much for working on this.

tute:
Yay! :tada: :balloon: :clap: 

Gonna go out in 4.4.

jyurek:
@betesh There are failures on Travis. Can you take a look at them and see if there's anything that jumps out at you? https://travis-ci.org/thoughtbot/paperclip/jobs/76525495

betesh:
@jyurek What was the goal of b8221c6b9f9e4c616973b859a8ad0b05f7511372?  If I understand correctly, it's to make the attachment URL use s3_doman_url by default if an s3_region has been defined.  Since s3_region is defined for AWS v2 in all the S3 specs (there's no other way to make them pass), we need either: 

1) if-then logic within each test to handle AWS v2 differently than v1 -- I just pushed a commit attempting this, we'll see if it works...

**OR**

2)  to consider those tests inapplicable to AWS v2, and just wrap the whole thing if an `if defined(::AWS)`

jyurek:
The goal was to get my local tests passing, but turns out I was wrong. I reverted and it's all good now. My bad. :)

betesh:
Just curious, do you mean tests in a rails app that uses paperclip, or the S3 Live specs?

jyurek:
I mean the Paperclip specs themselves. Though the failure I saw was in the feature specs, so technically it was an app that uses paperclip. But that's nitpicking.

nickrivadeneira:
@tute Looks like v4.4 is being skipped - will this officially be released in v5?

tute:
> @tute Looks like v4.4 is being skipped - will this officially be released in v5?

Yes. If you are on Rails 4.2 you can grab the latest v5 beta.

Issue #1326, opened 2013-08-29
A lot of temp files which are not explicitly unlinked


svetlins:
Hello,

On resizing uploaded images Paperclip creates a lot of Tempfiles (e.g. 6 styles = 13 Tempfiles) which are then not explicitly unlinked. On most Ruby implementations this is OK since the actual files are deleted when the Tempfiles instances are garbage collected. This is a problem though with JRuby since version 1.7.4 in which the behavior is changed so that files corresponding to Tempfiles are only deleted on process exit. Because of this our app started filling up /tmp since the upgrade to JRuby 1.7.4.

I would write an issue to JRuby about that but this piece of Ruby documentation discouraged me: http://www.ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html#label-Explicit+close

It basically says that both deleting behaviors are allowed and that it's good practice to explicitly unlink created temp files anyway.

Do you think it's possible to add explicit unlinking of all created temp files to Paperclip? That would really help users of latest JRubies!

Best,
Svetlin

maclover7:
Hi everybody! Is this still an issue for you in Paperclip; I know this issue is from approximately 2 years ago. If it is still an issue, can you please provide the code that's causing you the error? Thanks!

scomma:
This is still an issue for me. We use paperclip to store the attachments on S3, and the task that saves to S3 runs in a Resque job (which naturally executes in its own fork). Could this be a reason the garbage collector isn't kicking in?

topherhunt:
Looks like it's an issue for me too (Paperclip **4.2.0**); batch Paperclip operations tend to crash at the following error:

```
An Errno::EMFILE occurred in background at 2015-04-19 22:25:02 -0700 :

 Too many open files - identify -format '%wx%h,%[exif:orientation]' '/tmp/6f53a5e7b68ac7105b51677c0bc3426320150420-23382-17zy612.jpg[0]' 2>/dev/null
```

...which traces back to Paperclip and is discussed more in [this SO question](http://stackoverflow.com/questions/18393580/errnoemfile-too-many-open-files-with-local-images-on-create).

bluta:
also for me, (Paperclip 4.2.0, S3, uploading in Resque)
shame on me, I've only realized when it made our EngineYard util instance unusable - filling the 16G /tmp on the root partition.

jstorimer:
[...]

PR #1908, opened 2015-06-21
Regression fix: original file could be unlinked in post_process_style

tiegz:
https://github.com/thoughtbot/paperclip/commit/87cebf991ac49bb3f655e1e7d834c203217583dd added an `intermediate_files` local var in `Paperclip::Attachment#post_process_style` to collect temp files and then ensure they get unlinked at the end of `post_process_style()`.

But in this scenario [when a processor just returns `@file`]:

``` ruby
class Paperclip::Foo
  def make
    # ... do something with the file ...
    @file
  end
end

has_attached_file :image, {
  processors: [:foo],
  styles: {
    full: "200x200",
    thumb: "100x100"
  }
}
```

1) `post_process_style` is run for the `:full` style 
  * `Paperclip::Foo.make` is run on `:full`
  * the original file (aka `queued_for_write[:original]`) gets added to `intermediate_files`
  * when finished, files in `intermediate_files` are unlinked, including the original file

2) `post_process_style` is run for the `:thumb` style
  * `Paperclip::Foo.make` is run on `:thumb`  :bomb: :boom: `make` fails because the original file that was passed to it has been unlinked

NB I tried to write a spec for this, but it proved a little difficult. Hopefully this change makes enough sense when seen?

Thanks :sunrise:

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -522,7 +522,7 @@ module Paperclip
 
         @queued_for_write[name] = style.processors.inject(@queued_for_write[:original]) do |file, processor|
           file = Paperclip.processor(processor).make(file, style.processor_options, self)
-          intermediate_files << file
+          intermediate_files << file unless file == @queued_for_write[:original]
           file
         end
 
--- a/spec/paperclip/attachment_spec.rb
+++ b/spec/paperclip/attachment_spec.rb
@@ -699,9 +699,6 @@ describe Paperclip::Attachment do
 
     context "when assigned" do
       it "calls #make on all specified processors" do
-        Paperclip::Thumbnail.stubs(:make).with(any_parameters).returns(@file)
-        Paperclip::Test.stubs(:make).with(any_parameters).returns(@file)
-
         @dummy.avatar = @file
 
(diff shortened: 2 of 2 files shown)


jyurek:
Sounds good to me, but can you write a spec for it? Thanks.

tiegz:
@jyurek specs added!

dmitry:
Please make sure this will never happen again: https://github.com/thoughtbot/paperclip/issues/1326

tute:
Is this good for merge, @jyurek?

Fernan2:
I had my /tmp directory full of crap generated by paperclip... and this seems to be the solution, could you please merge it?

(that's not a trivial issue, I had to delete that crap manually because the server was about of running out of free disk)

dmitry:
@Fernan2 we've stuck the same issue with ext4, it has limitation of 4M files in on path. But we fixed it by changing the ENV variable for the resque: https://github.com/resque/resque/issues/1167

tute:
Thank you all!

tiegz:
Hmm, without merging this the regression still exists, and it's still possible for the `:original` to be unlinked before another style attempts to reference it. I thought my change was safe because `after_flush_writes` should clear the original -- is that not the case?

Issue #1476, opened 2014-03-05
validate_attachment_type_content_matcher: Short-curcuiting && means side effects do not happen; confusion around where errors are added


cofiem:
Similar, but not exactly the same, as #792 #1158 #1334 #161

Short-curcuiting && means side-effects of methods are not in effect
---------------------------
The return in `matches?` cannot be done as it is on line [38](https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/matchers/validate_attachment_content_type_matcher.rb#L38). This is because `&&` short-curcuits when a return value is false. 

A failure will call `failure_message`. Due to `failure_message` calling `accepted_types_and_failures` and `rejected_types_and_failures`, `@missing_rejected_types` and `@missing_allowed_types` must have values (`any?` is called on them). These two instance variables are set in the `allowed_types_allowed?` and `rejected_types_rejected?` methods.

The short-curcuit will mean that one or both of `allowed_types_allowed?` and `rejected_types_rejected?` are not called. The end result of all of this is the error discussed in #1334: `undefined method `any?' for nil:NilClass`

One possible fix is to replace line 37 and 38 with:
````ruby
rejected_t = rejected_types_rejected?
allowed_t = allowed_types_allowed?
@allowed_types && @rejected_types && allowed_t && rejected_t
````

Confusion around where errors are added
-----------------------------
Due to #1253, there is confusion about where to look for errors on `@subject` in the `type_allowed?` method.

Currently line [85](https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/matchers/validate_attachment_content_type_matcher.rb#L85) only checks ` @subject.errors[:"#{@attachment_name}_content_type"]`. It should also check ` @subject.errors[:"#{@attachment_name}"]`.

A possible fix is to replace line 85 with:
````ruby
content_type_any = @subject.errors[:"#{@attachment_name}_content_type"].any? { |item| item == 'is invalid' }
type_any = @subject.errors[:"#{@attachment_name}"].any? { |item| item == 'is invalid' }
!content_type_any && !type_any
````

maclover7:
Hi @cofiem ! Is this still an issue for you in Paperclip; I know this issue is from approximately 1 year ago. If it is still an issue, can you please provide the code that's causing you the error? Thanks!

cofiem:
[...]

PR #1910, opened 2015-06-23
Fix a nil error in failure_message of content type validation matcher

### LINKED ISSUE #1334: Strange Behaviour on validate_attachment_content_type
So, I was writing a test to validate the content_type of some attachments.

```ruby
  it { should validate_attachment_content_type(:image).allowing(
      "image/png", "image/jpg", "image/jpeg"
    ).rejecting(
      "text/plain", "text/html", "text/xml", "application/octet-stream", "application/exe"
    )
  }
```

And the output from the test was:

```
  2) Step should validate the content types allowed on attachment image
     Failure/Error: it { should validate_attachment_content_type(:image).allowing(
     NoMethodError:
       undefined method `any?' for nil:NilClass
     # ./spec/models/step_spec.rb:11:in `block (2 levels) in <top (required)>'
```

Then trying to fix that I discovered that if rejected_types_rejected? must be, for some reason I don't know, called first

So when I placed that code:

```ruby
class Paperclip::Shoulda::Matchers::ValidateAttachmentContentTypeMatcher
  def matches? subject
    @subject = subject
    @subject = @subject.new if @subject.class == Class
    @allowed_types && @rejected_types &&
    rejected_types_rejected? && allowed_types_allowed?
  end
end
```

The return from RSpec went as I expected:

```
  2) Step should validate the content types allowed on attachment image             
     Failure/Error: it { should validate_attachment_content_type(:image).allowing(
       Expected image:

       Accept content types: image/png, image/jpg, image/jpeg
         image/png, image/jpeg were rejected.

       Reject content types: text/plain, text/html, text/xml, application/octet-stre
am, application/exe
         All were rejected successfully.
```

I didn't created a Pull Request becouse that sounded really weird to me, so I decided just to open an issue.

Any clues on what might be going wrong? Anyone else can replicate the error there?

Thank you folks.

lime:
When a matcher uses both `allowing` and `rejecting`, and validation fails, `@missing_rejected_types` might still be `nil`. The validation itself works as intended, but calling `failure_message` raises due to `@missing_rejected_types.any?`.

Presumably, this fixes #1334. Possibly also #1476.

---

I haven't added any tests for this fix yet. Any suggestions on what tests we would want? The minimal case would be:
```
expect{ matcher.failure_message }.not_to raise_error
```
But that feels dumb. On the other hand, testing the failure message string for some specific pattern feels like overkill. Ideas?

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,7 @@ master:
 
 * README adjustments for Ruby beginners (add links, elucidate model in Quick Start)
 * Bugfix: Now it's possible to save images from URLs with special characters [#1932]
+* Fix a nil error in content type validation matcher [#1910]
 
 5.0.0.beta2 (2015-04-01):
 
--- a/lib/paperclip/matchers/validate_attachment_content_type_matcher.rb
+++ b/lib/paperclip/matchers/validate_attachment_content_type_matcher.rb
@@ -40,9 +40,9 @@ module Paperclip
 
         def failure_message
           "#{expected_attachment}\n".tap do |message|
-            message << accepted_types_and_failures
+            message << accepted_types_and_failures.to_s
             message << "\n\n" if @allowed_types.present? && @rejected_types.present?
-            message << rejected_types_and_failures
+            message << rejected_types_and_failures.to_s
           end
         end
 
@@ -55,7 +55,7 @@ module Paperclip
         def accepted_types_and_failures
           if @allowed_types.present?
             "Accept content types: #{@allowed_types.join(", ")}\n".tap do |message|
-              if @missing_allowed_types.any?
+              if @missing_allowed_types.present?
                 message << "  #{@missing_allowed_types.join(", ")} were rejected."
               else
                 message << "  All were accepted successfully."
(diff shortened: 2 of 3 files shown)


jyurek:
Can you add a test to demonstrate the error this may fix?

lime:
> Can you add a test to demonstrate the error this may fix?

Sure, sure. As I wrote above, I'm requesting advice on how extensively we want to test the output of `failure_message`. However, I guess it's easier to discuss that if we have some starting point.

lime:
Could I maybe get some feedback on this?

I rebased the commits on top of the current master and pushed them one by one. This should show that the fix still applies, and that `failure_message` breaks without it.

tute:
This is great, @lime. Can you please rebase again on top of latest master? And please add a note on the change to the `NEWS` file, and we'll merge. Thank you! 😃

lime:
Thanks for the update! I'll take a look at it tonight. :)

lime:
Rebased, sorry for the delay.

tute:
Thanks!!

PR #1916, opened 2015-07-01
add override_path_permissions option

### LINKED ISSUE #1865: Allow to set permissions of the created directory path
There's an ``override_file_permissions`` option to change the permissions of the file created, but not option to set the permissions of the path that is created – this prevents, for example, from a process running under a different user from deleting the file.

I suggest adding ``override_path_permissions`` option that will eventually get to ``/lib/paperclip/storage/filesystem.rb`` as an option in the ``FileUtils.mkdir_p`` call at line 38.

herregroen:
There's an `override_file_permissions` option to change the permissions of the file created, but not option to set the permissions of the path that is created. This prevents, for example, from a process running under a different user from deleting the file.

Used chmod only on newly created directories for similar functionality to override_file_permissions ( as opposed to the mode option in mkdir_p, which can not override the umask ).

Could potentially refactor to use the mode option but behaviour would vary depending on the active umask, which may not be desirable.

As a sidenote: tests will not pass on a shared directory within a VM due to inability to set permissions on directories.

[fixes #1865]

### DIFF
--- a/lib/paperclip/storage/filesystem.rb
+++ b/lib/paperclip/storage/filesystem.rb
@@ -35,7 +35,21 @@ module Paperclip
 
       def flush_writes #:nodoc:
         @queued_for_write.each do |style_name, file|
+          unless @options[:override_path_permissions] == false
+            dir  = File.dirname(path(style_name))
+            dirs = []
+            while not Dir.exists? dir
+              dirs << dir
+              dir = File.dirname(dir)
+            end
+          end
           FileUtils.mkdir_p(File.dirname(path(style_name)))
+          unless @options[:override_path_permissions] == false
+            resolved_chmod = @options[:override_path_permissions] || (0777 &~ File.umask)
+            dirs.each do |d|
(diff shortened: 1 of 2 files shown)


lib/paperclip/storage/filesystem.rb
```diff
@@ -35,7 +35,21 @@ def exists?(style_name = default_style)
 
       def flush_writes #:nodoc:
         @queued_for_write.each do |style_name, file|
+          unless @options[:override_path_permissions] == false
```
jyurek: This is a lot to be putting into this method. Can you extract these two sections into methods?

jyurek:
Looks pretty good, but I have a comment. Also, can you handle the Hound checks? Thanks!

tute:
> Looks pretty good, but I have a comment. Also, can you handle the Hound checks? Thanks!

ping @herregroen. Would you like to wrap up this PR? Thanks!

sidraval:

Issue #1884, opened 2015-06-03
Paperclip doesn't recognize SVG image data URIs unless they're base64-encoded


bjmc:
The regex for the [DataUriAdapter](https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/io_adapters/data_uri_adapter.rb#L4) explicitly requires base64 encoding:

    REGEXP = /\Adata:([-\w]+\/[-\w\+\.]+)?;base64,(.*)/m

However, there's no necessity to base64-encode SVG images, and it's [actually suggested not to.](https://css-tricks.com/probably-dont-base64-svg/) Because Paperclip fails to recognize valid data URIs of the form `data:image/svg+xml;utf8,` users are forced to base64-encode SVGs in order to upload them to Paperclip.  As a future enhancement, it would be nice if that restriction were removed.

jyurek:
Honestly, I'd never considered this, since this is typically used for binary formats, but you make a compelling argument. If you want to make this change and submit a PR, I'd be all for it. Thanks!

PR #1918, opened 2015-07-02
Can now handle data svg

duhruh:
This addresses #1884 and my own issue which is that svg's encoded in base64 then getting sent to the server were causing the following error:

```
no decode delegate for this image format `' @ error/constitute.c/ReadImage/501.
```

after some digging i found that svg's needed to be prepended with the 

``` xml
<?xml version="1.0" standalone="yes"?>
```

header in order to be valid. I claim to be no expert on this at all, this just comes from my initial research on the subject and that these changes fixed the issues I was experiencing, so any and all critiques are welcome. Keep up the awesome work thoughtbot you guys are some of the more badass rails devs :100:
# Edit:

 So my first attempt was pretty naive in regards to how i parsed the uri (though it still would have worked for most use cases), seems like data uris are more complex than i originally thought, according to [RFC-2397](https://tools.ietf.org/html/rfc2397) and [wiki](https://en.wikipedia.org/wiki/Data_URI_scheme#Format) a charset and data uri needs to be specified in the following format. 

```
data:[<MIME-type>][;charset=<encoding>][;base64],<data>
```

So you are allowed to define a charset and base64 encoding. My original attempt only allowed for one or the other. So now i've parsed out the charset and am now decoding the payload based on the given charset if provided. There's one thing i'm sketchy about which is the whole `force_encoding` thing, _we could just ignore the charset_ which seems to be what the browser does but i don't believe that is the best solution. Also I've made the assumption here that when a data:uri is present and not base64 encoded to url unescape the payload, which makes sense to me as an image not base64 encoded would have to be url escaped to be properly displayed in the browser anyway (which is probably outside the scope of this lib). So the point in all of this is to extract a payload that can be properly saved to a file making it a valid svg/png/pdf/whatever i believe this refactor makes a fair attempt at grabbing the relevant data and transcoding it without regressing. Though still I would appreciate any feedback you guys can give on it.


lib/paperclip/io_adapters/data_uri_adapter.rb
```diff
@@ -10,10 +12,74 @@ def initialize(target_uri)
     private
 
     def extract_target(uri)
-      data_uri_parts = uri.match(REGEXP) || []
-      StringIO.new(Base64.decode64(data_uri_parts[2] || ''))
+      @data_uri_parts = (uri.match(REGEXP) || []).to_a
+
+      @payload = get_payload_from_match
```
jyurek: There is a lot of code dealing with parsing the URI. It seems like extracting an object to represent the data URI and handle the parsing would be a good idea here. That would leave this class to handle the Adapter parts.

jyurek:
It looks like it could use a little extraction to clean it up, but I like where this is headed.

bjmc:
:+1: Thanks for working on this.

lib/paperclip/io_adapters/data_uri_adapter.rb
bjmc: For an example of what that might look like, you could check out [this parser](https://github.com/timsly/carrierwave-data-uri/blob/master/lib/carrierwave_data_uri/parser.rb) @timsly wrote for his carrierwave-data-uri plugin.

I don't know if there's a rivalry between Paperclip and Carrierwave, but it seems like there might be an opportunity to combine efforts on Data URI handling here.

rsmithlal:
Is this going to be merged in at some point? Seems a bit stale yes? I would like to use this SVG  functionality =)

jpdimond:
Agreed, I'd love to see this merged in! I had to use the hack of adding the <?xml version="1.0" standalone="yes"?> to the top of my svg string.

sidraval:

PR #1920, opened 2015-07-02
Update README.md

jspizziri:
Update documentation to include Paperclip + Capistrano + S3 scenario.


jyurek:
This doesn't seem like a useful bit to include in the README. The `rails_root` interpolation is clearly documented as to what it returns, so intending to use it for an S3 bucket is a bad idea.

PR #1922, opened 2015-07-03
[ci skip] Add Ubuntu install instructions

rodrigoargumedo:
ImageMagick has a Ubuntu Launchpad package available to install and ready to use. This will help Ubuntu users to find the correct version of ImageMagick right of the box. :+1:


tute:
Thank you!

PR #1937, opened 2015-07-16
Update README to illustrate Ruby 1.9 hash syntax

tylerwillingham:
There was a hearty mixture of hash rockets and the Ruby 1.9 hash syntax so I decided to just convert everything to the 1.9+ syntax


maclover7:
Can you please rebase from master? Thanks!

PR #1939, opened 2015-07-18
Fix specs broken by aruba changes

dgynn:
This PR fixes the specs broken by the deprecation of `check_file_presence` in the latest aruba. While I was in there I also cleaned up the other deprecation warnings from aruba, rspec, and active record.


dgynn:
I've also updated the travis config to allow travis's bundler caching to take effect and to run the allow_failures rubies after the primary rubies. It looks like this will allow travis to run about 50% faster.

Appraisals
```diff
@@ -1,14 +1,11 @@
 appraise "3.2" do
-  gem "rails", "~> 3.2.0"
-  gem "paperclip", :path => "../"
+  gem "rails", "~> 3.2.15"
```
tute: I'd leave this as `~> 3.2.0`, as bundler will anyway install latest.

tute:
Thank you very much for your work. Can you please address my only comment, squash the commits together into one with a good commit message, and rebase on top of latest master? I'll merge right after. Thank you again!

dgynn:
Thanks, @tute . Squashed, rebased, and back to green.

tute:
That's great, thank you! :clap: :+1:

Issue #909, opened 2012-06-07
Slow generation of url


DriesS:
We are trying to optimize our views and on a page where 40 pictures are loaded with the following code :

```
= image_tag(product.pictures.first.data.url(:gallery))
```

We have a load time of 840ms, if we change it to the following code :

```
= image_tag("http://bucketname.s3.amazonaws.com/products/#{product.pictures.first.id}/gallery.jpg?1325844462"
```

We become a load time of 220ms.

It means the interpolation of s3_path_url is very slow. Somebody else is expecting the same problems? For the moment I created a helper that generates my urls :

```
def picture_url(picture, style)
  "http://bucketname.s3.amazonaws.com/products/#{picture.id}/#{style}.jpg"
end
```

Only problem I have here is that the cache key is not there and also the extension not.

deanputney:
This problem still seems to be an issue. +1 for reopening.

lucianosousa:
I remember that when I had this problem, I just migrate from fog to aws-s3 and had a better performance, but, it's incomparable with carrierwave or any other solution I used.

Looks like when I paperclip request for a image they authenticate all the time on AWS, and other solutions store a login in cache... looks like this way the problem.

Issue #557, opened 2011-08-13
Slow interpolations


vbichkovsky:
I had some performance issues, and profiling revealed that most of request time was spent on attachment.url calls. I fetch 360 objects per request, and each of them has 5 URLs: 2 for original attachments and 3 for thumbnails. Time is spent mostly in String#gsub method, as it is called for each URL and each possible interpolation. Thus, the more interpolations you have (for example, S3 adds its own 4), the more time it takes to generate an URL. I managed to improve performance a bit by excluding tags, which are not used in the pattern:

``` ruby
    def self.interpolate pattern, *args
      tags = pattern.scan(/:[^\/]+/)
      tags.inject( pattern.dup ) do |result, tag|
        result.gsub(/#{tag}/) do |match|
          send( tag[1..-1], *args )
        end
      end
    end
```

But it is still not fast enough, so I'm using my own helper with hard-coded strings instead of Paperclip's methods:

``` ruby
"/system/#{attachment}s/#{obj.id}/..."
```

I think performance can be improved by caching strings with pre-interpolated unfrequently changing parts of URL (like :attachment, :style, etc.) and interpolating only what changes from model to model (:id, :filename, etc.).

PR #1888, opened 2015-06-07
Cache interpolator methods and reduce memory allocations

dgynn:
The `Paperclip::Interpolations` module generates a lot of objects. As noted in #1882 there is memory allocated to get the list of interpolation methods and then again when `gsub` is called.

This commit memoizes the array of interpolation methods and the key-to-symbol mappings. It then uses `gsub!` to manipulate the result only when a key matches. The interpolation method cache is reset if any new interpolation is added.

I've benchmarked it using a simple `Attachment` object configured to use :s3 and calling `my_attachment.url(:thumb)`. The cached version is 4.5x faster and uses about 85% less memory. This is separate from the improvements in #1873 (which were also great).

There are still some tunings that could be done on the individual interpolations, like freezing strings. This commit doesn't address those.

Benchmark IPS comparison
```
Benchmarking interpolaters
Calculating -------------------------------------
original interpolations
                       319.000  i/100ms
cached interpolations
                         1.539k i/100ms
-------------------------------------------------
original interpolations
                          3.143k (± 5.3%) i/s -     15.950k
cached interpolations
                         15.429k (± 5.5%) i/s -     76.950k

Comparison:
cached interpolations:    15429.4 i/s
original interpolations:     3143.1 i/s - 4.91x slower
```

MemoryProfiler before:
```
Total allocated: 71828 bytes (520 objects)
Total retained:  756 bytes (6 objects)

allocated memory by gem
-----------------------------------
     71788  paperclip/lib
        40  ruby-2.2.2/lib
```
Memory Profiler after
```
Total allocated: 9937 bytes (98 objects)
Total retained:  998 bytes (8 objects)

allocated memory by gem
-----------------------------------
      9728  paperclip/lib
       209  ruby-2.2.2/lib
```

PR #1940, opened 2015-07-18
Interpolate URLs using a string rather than a regex

stretchkennedy:
Several other users have had problems with the speed of `Paperclip::Interpolations.interpolate` or with `model.attachment.url` (#557, #909). `gsub` with a string is significantly faster than `gsub` with a regex, and changing to a string also prevents users from creating interpolations with ? wildcards in them.

See:
```
irb(main):001:0> require 'benchmark'
=> true
irb(main):002:0> puts Benchmark.measure{100_000.times{":foo/:bar".gsub(/:#{:bar}/, 'BAR')}}
  2.290000   0.030000   2.320000 (  2.425150)
=> nil
irb(main):003:0> puts Benchmark.measure{100_000.times{":foo/:bar".gsub(":#{:bar}", 'BAR')}}
  0.860000   0.010000   0.870000 (  1.001322)
=> nil
```

### DIFF
--- a/lib/paperclip/interpolations.rb
+++ b/lib/paperclip/interpolations.rb
@@ -30,7 +30,7 @@ module Paperclip
     def self.interpolate pattern, *args
       pattern = args.first.instance.send(pattern) if pattern.kind_of? Symbol
       all.reverse.inject(pattern) do |result, tag|
-        result.gsub(/:#{tag}/) do |match|
+        result.gsub(":#{tag}") do |match|
           send( tag, *args )
         end
       end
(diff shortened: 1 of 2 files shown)


dgynn:
You might be interested in #1888 which also does gsub against a string token as well as some other performance optimizations. I've tried you test case with a `:foo?` interpolator and that works too.

stretchkennedy:
I totally missed #1888. I'm fine with closing this PR, but it would be great to get the test case into the other one.

jyurek:
Ok, then. Seeing as this is a subset of what #1888 provides, I'll close this. Thanks for submitting it, though, @stretchkennedy.

Issue #1946, opened 2015-07-24
v4.3.0 - Rails 3.2 support broken by Paperclip::Schema::Statements.remove_attachment


andrewhubbs:
It appears that support for Rails 3.2 is not working in Paperclip 4.3.0. If this is not by mistake, what is the correct version of Paperclip to use that still supports Rails 3.2?

### Details:

`Paperclip::Schema::Statements.remove_attachment` calls `remove_column` with the method signature `remove_column(table_name, column_name, type = nil, options = {})`.

Rails 3.2 version of this method expects `remove_column(table_name, *column_names)`.

Result is an error like this because the column type is interpreted as an additional column to be dropped:
`PG::Error: ERROR:  column "string" of relation "model" does not exist`

tute:
Thank you for reporting! Closing in favor of discussion in PR.

PR #1947, opened 2015-07-27
Change remove column call to properly respect method signature for Rails 3 vs Rails 4

andrewhubbs:
This is a fix for issue #1946.

No test included. Was unsure how to properly test Rails version specific behavior. Happy to add a test with some guidance.

### DIFF
--- a/lib/paperclip/schema.rb
+++ b/lib/paperclip/schema.rb
@@ -40,7 +40,11 @@ module Paperclip
         attachment_names.each do |attachment_name|
           COLUMNS.each_pair do |column_name, column_type|
             column_options = options.merge(options[column_name.to_sym] || {})
-            remove_column(table_name, "#{attachment_name}_#{column_name}", column_type, column_options)
+            if ActiveRecord::VERSION::MAJOR >= 4
+              remove_column(table_name, "#{attachment_name}_#{column_name}", column_type, column_options)
+            else
+              remove_column(table_name, "#{attachment_name}_#{column_name}")
+            end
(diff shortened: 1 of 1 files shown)


maclover7:
Thanks for submitting this patch! :smiley: 

As far as testing this, you can try and stub out ```ActiveRecord::VERSION::MAJOR``` to equal 4 (or whatever value you'd like) by doing the following in an RSpec test: ```stub_const("ActiveRecord::VERSION::MAJOR", 4)```.

lib/paperclip/schema.rb
```diff
@@ -40,7 +40,11 @@ def remove_attachment(table_name, *attachment_names)
         attachment_names.each do |attachment_name|
           COLUMNS.each_pair do |column_name, column_type|
             column_options = options.merge(options[column_name.to_sym] || {})
-            remove_column(table_name, "#{attachment_name}_#{column_name}", column_type, column_options)
+            if ActiveRecord::VERSION::MAJOR >= 4
+              remove_column(table_name, "#{attachment_name}_#{column_name}", column_type, column_options)
```
tute: Is `column_type, column_options` needed here? Can we just remove it for any version of Rails?

tute: We don't need that, see 127e8797fb3c6daa2dbcd6763778d58b6645777f and 6ea7c26512434a93b598602fd6e959452a97969b. Will remove.

tute:
Thank you! Merged into master.

PR #1948, opened 2015-07-28
avoids comparing [] (from filename media-type guess) 

### LINKED ISSUE #1944: some file types have empty file types from extension, but have a `file` inferred type
e.g.
Content Type Spoof: Filename 150720_1.pcrd ( from Headers, [] from Extension), content type discovered from file command: application/zip. See documentation to allow this combination.

I think we don't have to worry about any disagreement between types when only one has been specified... pull request on the way

bwlang:
to inferred(from `file` command) or supplied types

if there are no media types from name, we should not try to compare them

fixes #1944

### DIFF
--- a/lib/paperclip/media_type_spoof_detector.rb
+++ b/lib/paperclip/media_type_spoof_detector.rb
@@ -30,7 +30,8 @@ module Paperclip
     end
 
     def media_type_mismatch?
-      supplied_type_mismatch? || calculated_type_mismatch?
+      #if there are any media types from the filename, check for mismatch with calculated and supplied types
+      media_types_from_name.size > 0 && (supplied_type_mismatch? || calculated_type_mismatch?)
     end
 
     def supplied_type_mismatch?


jyurek:
I think a better case would be that if there is a type you're going to regularly see, you should explicitly allow/expect it by registering that type, no? While the extension may not mean anything to `file` it could mean something to the user's machine, couldn't it?

What do you think here?

bwlang:
@jyurek 

This code is all server side... we have no way to know what a user's machine will do with any given type.
IMHO - we should catch the obviously bad stuff where content inferred type does not match name inferred type.  I don't think it makes sense for every user of this library to whitelist all the file types they might use but that file does not know about.

I don't think the failing tests have anything to do with this code... but let me know if you think otherwise and I'll investigate further.

tute:
Hi @bwlang, thanks for this. Is it still relevant? If so, can you add more context to the commit message, and regression tests? Thank you!

bwlang:
@tute 
Yes I think it's still relevant.
What else would you like to see here?

tute:
Hi! Thanks. I need to see:

1. What was your intended situation, and what was instead the behavior? How can I reproduce this?
2. A good commit message, explaining previous question. See also http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html. 
3. Add a regression test, so we make sure the bug (if it is so) doesn't reappear in future changes.

Does this make sense?

Thanks!

Tute.

tute:
ping @bwlang. Thank you!

bwlang:
I still plan to work on this... just have not found the time yet.

PR #1970, opened 2015-08-24
Incompatible aws versions

betesh:
Aws::S3::Object#exists? isn't defined in aws-sdk versions between 2.0.0 and 2.0.33

See https://github.com/aws/aws-sdk-ruby/releases/tag/v2.0.34


jyurek:
Merged this in. It makes the full suite take a LOOONG time, so we should look into that. But it's green. Thanks!

Issue #1973, opened 2015-08-26
Rails 5 Support


maclover7:
Not sure if any breaking changes that are upcoming in Rails 5 affect Paperclip, but it's important that the project is adapted if any changes are needed. @jyurek, will we be deprecating Rails 3.2 support when Rails 5 is released?

jyurek:
Yeah, I'd say we should. There are lots of Rails 3.2 projects out there, but we shouldn't hold things back because of them. And they can use older versions of Paperclip anyway.

Can you do a cursory check by adding a new Appraisal and seeing if there are any glaring breakages to get the ball rolling?

maclover7:
:+1:. Will send in a PR.

tute:
Thanks for this initiative @maclover7! Let's work on the PR, closing in favor of #1976!

PR #1976, opened 2015-08-28
Add Rails master appraisals in preparation for Rails 5

maclover7:
#1973 

Also reorganized Appraisals to make it easier to understand


Appraisals
```diff
@@ -1,29 +1,39 @@
...
+  gem "rails", "~> 4.2.0"
+  gem "aws-sdk", "~> 2.0"
 end
 
-appraise "4.2.awsv1" do
-  gem "rails", "~> 4.2.0"
+appraise "master.awsv1" do
+  gem 'rails', github: "rails/rails"
   gem "aws-sdk", "~> 1.5"
 end
```
tute: Do we need to take this combination into account? When is AWS v1 end of life?

Appraisals
```diff
@@ -1,29 +1,39 @@
-appraise "3.2" do
+appraise "3.2.awsv1" do
+  gem "rails", "~> 3.2.0"
+  gem "aws-sdk", "~> 1.5"
```
tute: Should we be testing with `~> 1.6`?

tute:
Thanks so much @maclover7! Seems like the red CI comes from `rails#master`? Should I restart the build (or you trigger it rebasing your commit on top of latest master)?

Thanks again!

maclover7:
Pushed up change that should fix dependency resolution problems - hopefully we can see whether the test suite passes on master now.

Appraisals
maclover7: Not sure about v1 support (this would be another big breaking change deprecation), but I can look into this.

tute:
> Not sure about v1 support (this would be another big breaking change deprecation), but I can look into this.

I don't know how AWS works. But if people are moving on to v2, as most Rails apps moved to 4, then it could be plausible to say "paperclip 5 only with AWS v2". But to be fair the code is already written, and I wouldn't expect it to change much, so it might not be worth it if it will bring unhappy users.

maclover7:
If we are going to deprecate AWS v1 support, I'd say we should include deprecation warnings in the next release out the door, to give time to upgrade.

tute:
> If we are going to deprecate AWS v1 support, I'd say we should include deprecation warnings in the next release out the door, to give time to upgrade.

:+1: :clap:

maclover7:
Should I open up a ticket on rails/rails for that "syntax error" that's breaking the rails master travis build? https://travis-ci.org/thoughtbot/paperclip/jobs/79552598#L237 @tute

tute:
> Should I open up a ticket on rails/rails for that "syntax error" that's breaking the rails master travis build? https://travis-ci.org/thoughtbot/paperclip/jobs/79552598#L237 @tute

It looks like that's Ruby >2 syntax. We should add an exception for Rails master/5 and Ruby 2 CI run (see for example https://github.com/doorkeeper-gem/doorkeeper/blob/master/.travis.yml#L16-L18).

maclover7:
@tute Can you check my .travis.yml syntax - updated it in last commit to try and stop 2.0 and rails master from running

.travis.yml
```diff
@@ -9,18 +9,25 @@ rvm:
...
+  - gemfiles/master.awsv1.gemfile
+  - gemfiles/master.awsv2.gemfile
 
 matrix:
   fast_finish: true
   allow_failures:
     - rvm: jruby-19mode
     - rvm: rbx-2
+  exclude:
+    - gemfile: master.awsv1.gemfile
```
tute: Do we need to add the `gemfiles/` path?

maclover7: yep, i think that's what the problem is. will squash down all of my commits before merging in :P

tute: > yep, i think that's what the problem is. will squash down all of my commits before merging in :P

Excellent, and rebase on top of latest master too. Thank you, @maclover7! :clap:

maclover7:
Rebased and consolidated --> just for maintenance of the code, is there any way to streamline our gemfiles and appraisals? Getting very cluttered now by testing 4 versions of Rails and 3 versions of aws-sdk. cc @jyurek

tute:
I've been restarting CI but we always get the same timeout. Makes me think there's actually something wrong with either rails or our code?

tute:
Rebased your branch on top of latest master. Right now, `master` is red for some cucumber steps, so if we get those it's not related with changes in this branch.

That said, latest CI run had the following failure that we'll need to address:

```
/home/travis/build/thoughtbot/paperclip/spec/spec_helper.rb:11:in `require': cannot load such file -- activerecord-import (LoadError)
```

Thanks!

.travis.yml
```diff
@@ -6,15 +6,29 @@ rvm:
...
   - gemfiles/3.2.awsv1.gemfile
+  - gemfiles/3.2.awsv2.0.gemfile
+  - gemfiles/3.2.awsv2.1.gemfile
   - gemfiles/4.1.awsv1.gemfile
+  - gemfiles/4.1.awsv2.0.gemfile
+  - gemfiles/4.1.awsv2.1.gemfile
   - gemfiles/4.2.awsv1.gemfile
+  - gemfiles/4.2.awsv2.0.gemfile
+  - gemfiles/4.2.awsv2.1.gemfile
+  - gemfiles/master.awsv1.gemfile
```
tute: I so want to get rid of AWS V1. :|

tute:
This PR is back to "normal" failures, to be worked on.

tute:
@maclover7 will rebase your branch, and make it point to 5 beta1, à la https://github.com/doorkeeper-gem/doorkeeper/pull/753, unless you'd rather do it yourself, or do it differently. Thanks!

tute:
Will start with smaller PRs removing from master unsupported versions of Rails (3.2 and 4.1), and AWS v1.

maclover7:
Rebased! Going to have a ton of time to work on this after tomorrow (yay school vacation!). Key stuff is to get out a 4.x release with deprecations, and then to move forward with 5.0.0, after Rails 5 is released.

tute:
That is great! You might want to rebase again after https://github.com/thoughtbot/paperclip/pull/2074, with smaller CI matrix.

I will now work on removing from master AWS v1 support.

.travis.yml
```diff
@@ -6,9 +6,21 @@ rvm:
 script: "bundle exec rake clean spec cucumber"
 
 gemfile:
-  - gemfiles/4.2.awsv2.1.gemfile
   - gemfiles/4.2.awsv2.0.gemfile
-  - gemfiles/4.2.awsv1.gemfile
+  - gemfiles/4.2.awsv2.1.gemfile
+  - gemfiles/master.awsv2.0.gemfile
+  - gemfiles/master.awsv2.1.gemfile
```
tute: Should we do `master` or `5.0.0.beta1`?

.travis.yml
```diff
@@ -6,9 +6,21 @@ rvm:
...
+
+matrix:
+  fast_finish: true
+  allow_failures:
+    - rvm: jruby-19mode
+    - rvm: rbx-2
```
tute: Why do we need to allow these failures?

gemfiles/4.2.awsv2.0.gemfile
```diff
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 
-gem "sqlite3", "~> 1.3.8", platforms: :ruby
+gem "sqlite3", "~> 1.3.8", :platforms => :ruby
```
tute: Shouldn't we keep the Ruby 2 hash syntax?

maclover7: This is autogenerated by Appraisals :(

tute:
It looks like we have some real failures now. Thank you, Jon!

Do you think we should point CI to 5.0.0beta1, to make sure we track more stable releases?

Thanks!

tute:
Rebased your branch with small changes Jon.

tute:
Did Rails 5 change Validators API? It looks like `AttachmentContentTypeValidator`'s `not` option is failing.

tute:
It looks like specs are written in a weird style. @jyurek, do you see any trouble with rewriting a spec like this? Does it test what was intended?

```diff
diff --git a/spec/paperclip/validators_spec.rb b/spec/paperclip/validators_spec.rb
index 398948d..d25ef56 100644
--- a/spec/paperclip/validators_spec.rb
+++ b/spec/paperclip/validators_spec.rb
@@ -19,10 +19,9 @@ describe Paperclip::Validators do
     end

     it 'prevents you from attaching a file that violates that validation' do
-      Dummy.class_eval{ validate(:name) { raise "DO NOT RUN THIS" } }
       dummy = Dummy.new(avatar: File.new(fixture_file("12k.png")))
+      dummy.valid?
       expect(dummy.errors.keys).to match_array [:avatar_content_type, :avatar, :avatar_file_size]
-      assert_raises(RuntimeError){ dummy.valid? }
     end
   end
```

tute:
@sgrif: is the behavior described in https://github.com/thoughtbot/paperclip/pull/1976#issuecomment-168777150 an expected ActiveRecord change? Thank you.

tute:
Hey, @maclover7! Will you continue working on this PR? Thank you!

tute:
The reason specs fail in Rails 5 seem to be that this validation block never runs: https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/validators.rb#L67-L69

```ruby
send(:"before_#{attribute}_post_process", :if => if_clause, :unless => unless_clause) do |*args|
  validator_class.new(options.dup).validate(self)
end
```

I can't spot why that is.

tute:
@rafaelfranca can you spot if a change in Rails 5 around callbacks (aside from the new option `ActiveSupport.halt_callback_chains_on_return_false`) might be messing something up here? Thank you!

rafaelfranca:
Not that I'm aware. The only change I remember is the way we internally execute a callback but that also changes in Rails 4.2 so you would be able to see the same problem in 4.2 if it was caused by that.

gemfiles/4.2.awsv2.0.gemfile
tute: Which makes sense for a testing library (https://github.com/thoughtbot/appraisal/issues/105#issuecomment-196308393). Thanks!

tute:
cc @kitop, thanks for pairing on this!

tute:
@kitop thank you very much for your help on this! I'd merge it as is, let me know if you'd like to change anything on the code or the commits before doing so.

PR #1985, opened 2015-09-02
Fixed s3 live specs

betesh:


tute:
Don't we need to have both versions, as we support AWS v1 and v2? You say you correct the specs, but they are green either way, how is that possible?

@jyurek @maclover7, until when do you think we should support AWS v1?

tute:
Thank you for your contributions! :clap:

betesh:
@tute The S3 live specs don't run on travis.  Certain environment variables need to be defined for them to run, and unless somebody volunteers to make their S3 credentials public, there's no way to set those environment variables on travis.

betesh:
I'm pretty sure with this PR, S3 live specs pass with both AWS v1 and AWS v2

tute:
Hi @betesh. Why do we need this change?

CI is red, can we get it green? Also, I'd solve Hound comments. Thank you for your work.

betesh:
The change was needed because these tests are broken.

These tests don't run on travis.

I didn't introduce the Hound violations--they were already there.

tute:
> The change was needed because these tests are broken. These tests don't run on travis.

So, this specs where red before but with your changes green?  How can I replicate that? Thanks!

betesh:
There are some environment variables that you need to set:

    S3_BUCKET=... S3_REGION=... AWS_ACCESS_KEY_ID=... AWS_SECRET_ACCESS_KEY=... rspec spec/paperclip/storage/s3_live_spec.rb

tute:
Can you please squash your commits together, and rebase against latest master? Also, if you can paste the exact command to run this spec I will appreciate it. We can add it as documentation wherever needed. Thank you! :+1:

betesh:
> Also, if you can paste the exact command to run this spec I will appreciate it. We can add it as documentation wherever needed. 

There is no exact command.  You need to fill in the blanks above with your S3 credentials.

tute:
@betesh thanks for your work and help. Indeed the syntax in the file is old, but when I set up the command with my AWS credentials in latest master, surprisingly, it doesn't fail. Why would that be?

tute:
```
paperclip(master)$ S3_BUCKET=... S3_REGION=us-east-1 AWS_ACCESS_KEY_ID=... AWS_SECRET_ACCESS_KEY=... be rspec spec/paperclip/storage/s3_live_spec.rb
Testing against version 4.2.6

Randomized with seed 8364
...........

Finished in 6 seconds (files took 0.754 seconds to load)
11 examples, 0 failures

Randomized with seed 8364
```

tute:
I perform changes to the file meaning to break it, but it does not. Before fixing it, I want to see it broken. How can we test for that?

betesh:
@tute These tests aren't particularly thorough.  It may be that they aren't exercising enough of the S3 features to make all the setup code actually used.

I am surprised by this too.

tute:
Merged into master. Thank you for your help!

Issue #1974, opened 2015-08-27
Default S3 permissions trigger Aws::S3::Errors::InvalidArgument from AWS SDK


camhine:
The default s3_permissions (:public_read) cause an Aws::S3::Errors::InvalidArgument exception when trying to `save` a model with an attached file.

Somewhere along the line support for :public_read has been dropped for 'public-read'.

Setting s3_permissions: 'public-read' in the config fixes the problem - 'public-read' should probably be the default going forward.

jyurek:
What version of Paperclip and the aws-sdk are you using? I've just pulled in a PR that allows for aws-sdk v2 compatibility, and which also changes the default when using aws-sdk v2 to `"public-read"`.

camhine:
I'm using paperclip from git, revision: ece46f46e9b9c327ee65402f6510557f1fca2ed1

aws-sdk is version 2.0.48

I've just noticed that aws-sdk-v1 (1.65.0) is also in my Gemfile.lock. Maybe the version 2 default is being clobbered by the version 1 default?

Broken paperclip initializer:

```
Paperclip::Attachment.default_options.merge!(
  bucket: ENV['AWS_BUCKET'],
	# s3_permissions: 'public-read',
  s3_region: ENV['AWS_REGION'],
  storage: :s3
)
```

Uncommenting that `s3_permissions` fixes the issue.

Backtrace:

```
Aws::S3::Errors::InvalidArgument - :
  aws-sdk-core (2.0.48) lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call'
  aws-sdk-core (2.0.48) lib/aws-sdk-core/plugins/s3_sse_cpk.rb:18:in `call'
  aws-sdk-core (2.0.48) lib/seahorse/client/plugins/param_conversion.rb:22:in `call'
  aws-sdk-core (2.0.48) lib/aws-sdk-core/plugins/response_paging.rb:10:in `call'
  aws-sdk-core (2.0.48) lib/seahorse/client/plugins/response_target.rb:18:in `call'
  aws-sdk-core (2.0.48) lib/seahorse/client/request.rb:70:in `send_request'
  aws-sdk-core (2.0.48) lib/seahorse/client/base.rb:216:in `block (2 levels) in define_operation_methods'
  aws-sdk-resources (2.0.48) lib/aws-sdk-resources/services/s3/file_uploader.rb:42:in `block in put_object'
[...]

PR #1986, opened 2015-09-02
defualt permission should assume aws v2 if both aws versions are in use

betesh:


tute:
These conditionals would not exist if we had two different files, one for each branch. This PR doesn't change behavior or simplify the code, so I won't merge it for now. I am very interested in simplifying this into two files: `s3-aws1.rb`,  and `s3-aws2.rb`. We can merge that PR if you'd like to work on it!

Thank you very much for your contribution.

betesh:
**This is a fix to code that has already been merged into master.**

@tute I assume from your later comment https://github.com/thoughtbot/paperclip/issues/1974#issuecomment-138757919 that you do plan on merging this after all?

The question of separate classes was brought up already in the discussion on the PR that introduced AWS v2 support: https://github.com/thoughtbot/paperclip/pull/1903#issuecomment-113193016

tute:
Thanks! All done then. Where did you read about the upcoming deprecation of v1? I couldn't find docs related to that.

betesh:
@tute, I didn't say anything about deprecation.

Are you merging this?

tute:
> @tute, I meant that eventually paperclip will want to deprecate AWS v1 support. I was assuming (correct me if I'm wrong) that once paperclip publishes a gem that supports both, we'll want to deprecate AWS v1 support a few versions later. (Reading that comment again now, I was also confused by what I meant.)

We have a specific issue to discuss wether AWS v1 should be deprecated: https://github.com/thoughtbot/paperclip/issues/1997.

betesh:
Thank you for pointing me to that ticket.

However, it has nothing to do with whether this PR should be merged.  There is a bug in master.  It should be fixed before we publish a gem.

betesh:
@tute I still don't understand why this was closed.  I have rebased it onto the latest master, where it is still an issue

tute:
I just came from vacations, will look into it soon.

tute:
Hi @betesh! I can't reopen the PR, because the branch seems to be gone from your repo. Want to pusblish it so we reopen?

But before, why is this change necessary? Thank you!

betesh:
AWS v1: 'public_read'
AWS v2: 'public-read'

If you have both in you Gemfile, we use v2 but we use public_read, which is inconsistent and means that the default configuration is broken.

PR #1991, opened 2015-09-05
Prepare release for v5.0.0

tute:
There are two backward incompatible changes.

cc @jyurek @iwz

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,4 @@
-master:
+5.0.0 (2015-??-??):
 
 * Improvement: Paperclip now supports aws-sdk v2
   @betesh, @davetchen,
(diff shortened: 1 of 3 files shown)


maclover7:
Should we hold off on releasing this until Rails 5 lands, I think we are planning on dropping support for 3.2 as apart of that release?

iwz:
There was a fix for Rails 3.2 I'd like to get into a  release. Maybe we can
cut from there?

tute:
> There was a fix for Rails 3.2 I'd like to get into a release. Maybe we can cut from there?

Sounds good. master got so many changes, including two backward incompatible ones. We could do a patch release with your commit. What is its SHA?

tute:
> I think we are planning on dropping support for 3.2 as apart of that release?

Yes.

> Should we hold off on releasing this until Rails 5 lands

We could hold off the backwards incompatible changes, but a minor version with everything else would be healthy I think. Let's ask @jyurek!

jyurek:
I think a minor version without the backwards incompatible changes are probably best right now, and we can discuss things like the MD5->SHA change.

iwz:
@tute the SHA I'm looking for is c740fb171fe2f88c60b999d2a1c2122f2b8f43e9

tute:
After we release https://github.com/thoughtbot/paperclip/pull/1995, I'll merge master into it, and rebase this branch when we decide it's time to do the major release. Thank you for your input.

greggawatt:
It seems like #1995 was merged in, what's the status on this?

tute:
We'll release the major version after Rails 5 is out. The commit that fixed a Rails 3.2 regression is already out.

PR #1993, opened 2015-09-07
Don't write original if it wasn't reprocessed

### LINKED ISSUE #1055: #reprocess! original images lost
I don't have a specific error narrowed down yet but I thought I should start the discussion.

In the midst of reprocessing several thousand images, I lost about 300-400 of them.  I was using a resque job to iterate through my ActiveRecord objects and spinning off delayed_paperclip jobs to reprocess each one.  

The odd part about this is that when they are 'lost', they still exist as a file in the original style but it's just a 0kb size file without any other styles processed.  I was under the impression that reprocess wouldn't touch the original file.  

Other pertinent information is that this is with the latest version of paperclip (3.3.0) and the images are being stored on S3

alexandrz:
If only_process list is not empty, but it doesn't contain :original style, original file hasn't been reprocessed and it's not needed to rewrite/reupload it.

Fixes #1055 again

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -236,7 +236,9 @@ module Paperclip
     # the instance's errors and returns false, cancelling the save.
     def save
       flush_deletes unless @options[:keep_old_files]
-      @queued_for_write.except!(:original) unless @options[:only_process].empty? || @options[:only_process].include?(:original)
+      if @options[:only_process].any? && !@options[:only_process].include?(:original)
+        @queued_for_write.except!(:original)
+      end
       flush_writes
       @dirty = false
       true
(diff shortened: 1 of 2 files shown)


tute:
Does this fix the issue? If so, why? Can you add that context to the commit message?
Can you add a regression test?
And, if you can address the hound comment, I agree with it: line is too long! :smile: 

Thank you very much. :+1:

alexandrz:
I am not sure that this solution is the best possible, but it fixed the issue for us and we needed to fix it asap. Maybe someone who knows the codebase better than me could offer another way of archiving it.

What this code does is it prevents saving of original image if it wasn't reprocessed. When reprocess! method is called only for some styles and original is not included in the list.

jyurek:
Can you please add a test for this to make sure it doesn't regress in the future?

alexandrz:
Added tests

tute:
Thanks for your work, @alexandrz! Merged into master. 🎉

PR #1996, opened 2015-09-09
Remove locales from main repository

tute:
Extract another repository only concerned with locales:
https://github.com/thoughtbot/paperclip-i18n

A user of paperclip might choose to add `paperclip-i18n` as a dependency
for their application, or copy over the related locale file into their
project.

This is a first step in splitting some parts of the codebase into other
projects. Simple and relevant contributions like new translations might
get lost in the shuffle of dozens of PRs awaiting for approval, that
take time to review and merge.

Locales are a simple enough problem to be handled in an external
repository, where tests for consistency across translations run in less
than a second, and PRs can almost trivially be merged in.

cc @yurek @sikachu @maclover7


maclover7:
:+1:. Is this a "breaking change"? Technically functionality is being removed (even though it's still available elsewhere). If so, that means this needs to be on hold until next major version (5.0.0).

Also, not sure if any tests that live here in Paperclip rely on localizations, but we should make sure we have proper test coverage with this move.

tute:
> :+1:. Is this a "breaking change"? Technically functionality is being removed (even though it's still available elsewhere). If so, that means this needs to be on hold until next major version (5.0.0).

It is a backwards incompatible change, because users potentially need to change their apps for incorporating this commit. So we need to hold it off indeed. Thanks!

> Also, not sure if any tests that live here in Paperclip rely on localizations, but we should make sure we have proper test coverage with this move.

`paperclip` runs tests against the `en` dictionary, so that's covered. And `paperclip-i18n` uses `i18n-spec` to make sure the translations stay in sync. So we have gained coverage with this split, because `paperclip` doesn't use `i18n-spec` as of now, and locales can get out of sync.

Will add a `v5` tag to this PR.

tute:
@jyurek will merge after your "ok". master is already backwards incompatible so it should be fine.

Issue #1997, opened 2015-09-09
Deprecate aws-sdk v1


maclover7:
As said by @tute, we need to deprecate the use of aws-sdk v1, as major version 2 has been released (currently at v2.1.19).

Steps we need to follow for this:
- Release a minor version which includes a deprecation warning if the user is still on aws-sdk v1
- Release a major version which includes the actual deprecation of aws-sdk v1 (this is a breaking change)

tute:
For context, I'd deprecate it if Amazon is deprecating it. Is it? If so, when? If not, it's an arguable step to take.

jyurek:
Agreed. I'd only want to deprecate it if Amazon is deprecating it.

maclover7:
https://github.com/aws/aws-sdk-ruby/issues/924

tute:
Thanks! It sounds like Rails 5 will come before they come up with a decision anyway.

maclover7:
yep, so let's have Rails 5 + dropping Rails 3.2 + any other backwards incompatible changes as apart of Paperclip 5.0.0. Should we release a minor version of Paperclip that includes deprecation warnings?

tute:
People on Rails 2.3.x may use paperclip 2.7, people on Rails 3 may use paperclip 3, and so on. I don't think we need deprecation warnings for Rails.

If there are other more subtle deprecations or changes of functionality that could be surprising, I'm all for warnings.

tute:
I'll bit the bullet and call for a removal of AWS V1 for paperclip 5. trevorrowe says in https://github.com/aws/aws-sdk-ruby/issues/924#issuecomment-140182581:

> we took a lot of effort to ensure that version 1 and version 2 could be used in the same project allowing users to upgrade at gradually and at will

v1 users can continue to use paperclip 4.x.

maclover7:
Closing in favor of #2049.

Issue #1955, opened 2015-08-05
Temp file with length of 0 raises NotIdentifiedByImageMagickError, could be better handled


dpritchett:
cross post from #1223:

> @leeizamo I just had the same error crop up for me but the only time it happened was when I was reprocess!ing and image off of S3. Turns out the image dimensions check first redownloads the image from S3 into /tmp and then tries to check its size using local ImageMagick tools (file -b --mime /tmp/filenamehere. For reasons unkown to me the tmp file was there but its length was 0.

> On a lark I tried messing with a few different versions of aws-sdk.

> Pinning my aws-sdk version to ~> 1.6 as mentioned in a comment in Paperclip's S3 class seems to have fixed it.

This error seems to be a side effect of a bad `aws-sdk` version.  Given that paperclip doesn't force an aws sdk version we could maybe consider insisting of a specific version (you probably already have a reason for not doing that) or at least checking for cases when the temp file exists but has a length of 0 and throwing a more specific error.

Thanks!

tute:
> Given that paperclip doesn't force an aws sdk version we could maybe consider insisting of a specific version

I am all for it, it seems like many difficulties in paperclip arise from that. What would be a sensible version to lock paperclip to? @jyurek do you see any problem with this approach?

jyurek:
I'm all for putting version restrictions on it, except for when I do, it seems that someone else has a problem with the way it got done. I don't really know what to do here.

dpritchett:
Maybe we can duck type all of the requirements from AWS rather than
actually locking a version; could even rescue the common AWS errors and
throw out some "try version XYZ!" error messages.

tute:
Closing in favor of discussion in PR.

PR #2005, opened 2015-09-18
Up the aws-sdk v1 requiremet to >= 1.64.0

jyurek:
@tute This should address the concerns of making sure we don't have 0-length files anymore. I don't know if this is enough to cause a minor version bump though. Thoughts?

### DIFF
--- a/paperclip.gemspec
+++ b/paperclip.gemspec
@@ -35,7 +35,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency('rspec', '~> 3.0')
   s.add_development_dependency('appraisal')
   s.add_development_dependency('mocha')
-  s.add_development_dependency('aws-sdk', '>= 1.5.7', "< 3.0", *((0..33).to_a.collect{ |release_number| "!= 2.0.#{release_number}" }))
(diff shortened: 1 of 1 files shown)


tute:
Quoting your in https://github.com/thoughtbot/paperclip/issues/1955#issuecomment-141469883:

> I'm all for putting version restrictions on it, except for when I do, it seems that someone else has a problem with the way it got done. I don't really know what to do here.

What are those problems? Are they in this issue tracker?

jyurek:
They are, as they're things people would report, but they're scattered. I think it's probably just best to do it, bite the bullet, and point to this as a reason if anyone has issues.

tute:
Your call on merge then. To me this smells like a good reason to merge in a deprecation notice and call version 1 just buggy.

tute:
Should we close this PR, and start a new "AWS V1" deprecation notice? See https://github.com/thoughtbot/paperclip/issues/1997#issuecomment-155987648

tute:
AWS V1 has been removed: https://github.com/thoughtbot/paperclip/pull/2075. Closing this PR.

PR #2023, opened 2015-10-12
Update instance of Proc.

gfvcastro:


tute:
Thank you for your input, @gfvcastro. I won't merge it in though, because this class has the highest churn in paperclip and I don't want to pollute the git history with styling changes. If we touch it, I'd rather extract objects from it, to split v1 and v2 behavior into different classes for example.
Thank you very much anyway!

PR #2028, opened 2015-10-19
make style optional in copy_to_local_file

kurtzur:
I recently had to use `copy_to_local_file` and was frustrated by the fact that `style` was a required parameter. I had to dig into `Attachment::default_options` to realize that `:original` was what I needed to pass.
I changed `copy_to_local_file` so that the style falls back to `default_style` when not provided. This matches the behavior of existing methods like `expiring_url` and `s3_object` (in `module S3`).

### DIFF
--- a/lib/paperclip/io_adapters/attachment_adapter.rb
+++ b/lib/paperclip/io_adapters/attachment_adapter.rb
@@ -24,7 +24,7 @@ module Paperclip
       if source.staged?
         FileUtils.cp(source.staged_path(@style), destination.path)
       else
-        source.copy_to_local_file(@style, destination.path)
+        source.copy_to_local_file(destination.path, @style)
       end
       destination
     end
--- a/lib/paperclip/storage/filesystem.rb
+++ b/lib/paperclip/storage/filesystem.rb
@@ -19,7 +19,7 @@ module Paperclip
     #   saved by paperclip. If you set this to an explicit octal value (0755, 0644, etc) then
     #   that value will be used to set the permissions for an uploaded file. The default is 0666.
     #   If you set :override_file_permissions to false, the chmod will be skipped. This allows
-    #   you to use paperclip on filesystems that don't understand unix file permissions, and has the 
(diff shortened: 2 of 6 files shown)


lib/paperclip/storage/filesystem.rb
```diff
@@ -19,7 +19,7 @@ module Storage
     #   saved by paperclip. If you set this to an explicit octal value (0755, 0644, etc) then
     #   that value will be used to set the permissions for an uploaded file. The default is 0666.
     #   If you set :override_file_permissions to false, the chmod will be skipped. This allows
-    #   you to use paperclip on filesystems that don't understand unix file permissions, and has the 
+    #   you to use paperclip on filesystems that don't understand unix file permissions, and has the
```
kurtzur: :no_mouth:

lib/paperclip/io_adapters/attachment_adapter.rb
```diff
@@ -24,7 +24,7 @@ def copy_to_tempfile(source)
       if source.staged?
         FileUtils.cp(source.staged_path(@style), destination.path)
       else
-        source.copy_to_local_file(@style, destination.path)
+        source.copy_to_local_file(destination.path, @style)
```
tute: Do we need this change?

tute: > Do we need this change?

Yes, I see why now.

lib/paperclip/storage/s3.rb
```diff
@@ -446,10 +446,10 @@ def flush_deletes #:nodoc:
         @queued_for_delete = []
       end
 
-      def copy_to_local_file(style, local_dest_path)
+      def copy_to_local_file(local_dest_path, style_name = default_style)
```
tute: We don't need to switch the order of parameters, can we leave them as they were? Thank you.

tute:
I'd rather have this method require the style and keep it explicit. If you think we can improve the documentation, or something else, please let me know. Thank you for sharing your work!

PR #2031, opened 2015-10-21
Return as json?

SylarRuby:
Is it possible to return the `avatar.url` as json?

In the rails console I can `user.profile.avatar.url` and it gives me the url but how to return that as json? Need it for reactjs.


maclover7:
Thanks for your PR! Can you check out the commit history -- I think something may have gone wrong with a rebase. Also, can you resolve the Hound CI comments? Thanks! :)

tute:
This PR goes from master to v4.3. We can't do that merge, it would make v4.3 backward incompatible.

PR #2033, opened 2015-10-23
Support ruby3

sachin21:
# WHAT
Supported ruby3.
Ruby3 String will not be able to use the `<<` without `.dup`.

# REF
- https://bugs.ruby-lang.org/issues/11473 (English)
- http://d.hatena.ne.jp/ku-ma-me/20151004/p1 (Japanese)


sachin21:
@houndci Thank you for your comments :smile: 
I fixed issues of rubocop on this PR and squashed commits :+1: 
But I didn't resolve issues of Rubocop that exists already.

lib/paperclip/matchers/validate_attachment_content_type_matcher.rb
```diff
@@ -65,7 +65,7 @@ def accepted_types_and_failures
         end
         def rejected_types_and_failures
           if @rejected_types.present?
-            "Reject content types: #{@rejected_types.join(", ")}\n".tap do |message|
+            "Reject content types: #{@rejected_types.join(", ")}\n".dup.tap do |message|
```
sachin21: @houndci I thought paperclip is not use the rubocop...
sorry. I fixed that

lib/paperclip/matchers/validate_attachment_content_type_matcher.rb
```diff
@@ -54,7 +54,7 @@ def description
 
         def accepted_types_and_failures
           if @allowed_types.present?
-            "Accept content types: #{@allowed_types.join(", ")}\n".tap do |message|
+            "Accept content types: #{@allowed_types.join(", ")}\n".dup.tap do |message|
```
sachin21: too ref: https://github.com/thoughtbot/paperclip/pull/2033#discussion_r42825289

sachin21:
@houndci Would you mind to merge this PR?

sachin21:
omg
I found that out Mr. houndci is bot.
@tute How about this PR?

tute:
Hey, thanks so much! But until we don't have Ruby 3 in CI, we can't, as we don't have regression tests. Does Travis already support Ruby 3?

sachin21:
hmm, That is certainly true.
We'll wait once until it's supported. so May I stay this PR? until it's supported.
At that time remind to you :smile:

tute:
Yep! Pending a version of ruby 3 in Travis. :+1:

sachin21:
> Ruby 2.3 has a feature to help transition to the frozen string change of Ruby 3. Take a look at the section Frozen String Literals on http://nithinbekal.com/posts/ruby-2-3-features/. It looks like we can add a comment # frozen_string_literal: true to enable the frozen string functionality of Ruby 3. Ruby 2.3 does not have an official build yet so this may be a bit premature. I think we should add in the comment and fix the errors that occur because of the change.

https://github.com/bbatsov/rubocop/pull/2328#issuecomment-163628935

I think this idea is good! How about this latest commits?

@tute

tute:
Thanks! Two questions:

* Can we enable this option automatically on every paperclip file, without adding a comment to each one of them?
* Instead of mutating the `path` string, can we build it in such a way that we can work on it as a frozen string?

Thanks again!

sachin21:
> Can we enable this option automatically on every paperclip file, without adding a comment to each one of them?

No sir, We can't do that now. But when ruby3 is released, We can remove inline comment(for frozen string literal) in all *.rb files.
so let's wait the ruby3 version :smile: 
Next, 
> Instead of mutating the path string, can we build it in such a way that we can work on it as a frozen string?

Uhh, Do you mean this? https://github.com/sachin21/paperclip/commit/2174c915507587d8f7f9f74457c9d7fabd025201#diff-2590d28581541e7cccb81df6b758e191L5

Then, I’d like to propose this `# frozen_string_literal: false` and remove the dup method.
So Travis CI will be passed testing.

bdewater:
Only string literals like `"foo"` are frozen by the comment. You could use `String.new("foo")` and mutate it. Call `#freeze` yourself if you want to have frozen strings everywhere.

sachin21:
@bdewater ok I replaced to String.new instead of quotes :smile: 
and...
@tute how about [this](https://github.com/thoughtbot/paperclip/pull/2033#issuecomment-167507315) latest comment for u?

tute:
> @tute how about this latest comment for u?

Wait for a version of ruby 3 makes sense to me, so we can test without the comments, and current solution for the path string (without dup) too.

Thank you!

sachin21:
I understand. I'll close this pull-request :smile:

PR #2043, opened 2015-11-10
Typo

LongMan:


tute:
Thanks! I don't know why would Travis fail, restarted it and will see it's result (and wait for it to be green) before merging any commit into master.

tute:
Fails on cucumber specs, it looks like it's because `tmp/aruba/testapp/` doesn't have a DB to connect to. Will open up a new issue. Thanks!

tute:
Thanks! I decided to merge, and if CI is red in master this won't change it. :)

PR #2046, opened 2015-11-10
Fixes #1804 - Reprocessing a single style re saves the original

### LINKED ISSUE #1804: Reprocessing a single style re saves the original
It appears that calling ```attachment.reproces(:my_style)``` re-saves the original image as well.
This results in an extra upload of the original file to S3 (if you use it), causing extra rountrip / delay (and potentially a new file if your interpolations include :updated_at.

In reality reprocessing a single style should not update the original record I believe, as that has not changed?

This is more apparent if you are using the :only_process option and delayed_paperclip with split processing, eg:

````
    has_attached_file :file,
                      storage: :s3,
                      styles: lambda { |_a| 
                                            {
                                              original: { geometry: "2880x2880>" },
                                              thumb: { geometry: "260x260#" }
                                            }                                            
                      processors: [:rotator, :thumbnail, :paperclip_optimizer],
                      only_process: [:original]

  process_in_background :file, only_process: [:thumb]
````

In the scenario above, I would expect:
* paperclip to process the original on upload
* delayed_paperclip to process :thumb asynchronously (we use sidekiq)

What happens is:
* paperclip processes the original on upload - correct
* delayed_paperclip calls attachment.reprocess!(:thumb) from a sidekiq job - correct
* paperclip re-saves the original again and reprocesses the :thumb - incorrect, only the thumb should be saved / uploaded to S3

the culprit seems to be this bit of code:
https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/attachment.rb#L426

When calling reprocess! passing only certain styles, the original is still @queued_for_write.

I'm happy to try to open a PR for this, but I would like to know if you guys have any suggestion on how to approach this, shall we skip the following if :only_process does not include the original?

https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/attachment.rb#L107
https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/attachment.rb#L109

Any input is appreciated.

MarkMurphy:
All tests are currently passing but I haven't written any new tests to cover this new scenario. Hoping someone else is up to the task :+1:

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -334,6 +334,16 @@ module Paperclip
       saved_preserve_files, @options[:preserve_files] = @options[:preserve_files], true
       begin
         assign(self)
+
+        # Unset @queued_for_write[:original] if it wasn't intended to be
+        # reprocessed. This seems a bit awkward but @queued_for_write[:original]
+        # is relied on to be present thought this code. Assumptions were made that
+        # we can't really back out of without a much more significant refacor.
+        #
+        # See: https://github.com/thoughtbot/paperclip/issues/1804
+        # and possibly related: https://github.com/thoughtbot/paperclip/issues/1936
+        @queued_for_write.delete(:original) unless process_style?(:original, only_process)
+
         save
         instance.save
       rescue Errno::EACCES => e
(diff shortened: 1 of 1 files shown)


dmitry:
@MarkMurphy can you please rebase this PR?

tute:
Fixed through https://github.com/thoughtbot/paperclip/pull/1993. Thank you for your work!

PR #2047, opened 2015-11-11
Fix typo

zzet:


tute:
This changes this intended structure:

```ruby
config.paperclip_defaults = {
  storage: :fog,
  fog_credentials: {
    provider: "Local",
  	local_root: "#{Rails.root}/public"
  },
  fog_directory: "",
  fog_host: "localhost"
}
```

Into this incorrect one:

```ruby
config.paperclip_defaults = {
  storage: :fog,
  fog_credentials: {
    provider: "Local",
    local_root: "#{Rails.root}/public}",
    fog_directory: "",
    fog_host: "localhost"
  }
```

It's correct as it is in current master.

It would be clearer if instead of a long line it was written this way, I welcome a PR adding new lines and white space.

Thanks!

Issue #2048, opened 2015-11-12
FIX CI in master


tute:
It looks like the dummy app for cucumber features doesn't have a database, and those specs are erroring out. This should be fixed, so that master is green in Travis again.

tute:
Argh, busted caches maybe?

dgynn:
I think the dummy app is using the config/database.yml rather than the spec/database.yml due to changes in aruba 0.10.x but I wasn't able to track it down any further.

PR #2051, opened 2015-11-16
Resolve broken CI tests due to upstream gem changes

dgynn:
There was a change in `aruba` that is breaking the cucumber tests. This PR doesn't fix the root problem, it just locks down the version to make it work. This partially addresses #2048 

The `fog` gem has been making updates to loosen dependencies on net-ssh (see fog/fog-core/pull/166 or fog/fog/commit/ebfaa0c97a387eeb18c5c797aaa33a1596e46676). Rather than explicitly requiring net-ssh this PR changes the development dependencies to the 2 providers that are tested.

### DIFF
--- a/paperclip.gemspec
+++ b/paperclip.gemspec
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency('aws-sdk', '>= 1.5.7', "< 3.0", *((0..33).to_a.collect{ |release_number| "!= 2.0.#{release_number}" }))
   s.add_development_dependency('bourne')
   s.add_development_dependency('cucumber', '~> 1.3.18')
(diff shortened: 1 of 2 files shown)


tute:
Thank you! :clap: :smiley:

geemus:
Sorry for the pains, trying our best to still support those stuck on 1.9, but it is getting more awkward/challenging all the time.

tute:
Thanks for reaching out and for your work, @geemus.

geemus:
Sure thing. Just let me know if there is anything I can do, trying to help sort it out but the number of moving pieces makes it pretty hard to have high confidence/certainty that the problems aren't just getting shuffled around sometimes.

PR #2054, opened 2015-11-16
Cache interpolator methods and reduce memory allocations - rebased

dgynn:
This is a rebased PR of #1888 in case that one does not merge cleanly to master (although it should).


dgynn:
Closing since v4.3 was merged in directly.

PR #2058, opened 2015-11-19
Add an interpolation for "base_class"

kddeisz:
When using Paperclip and Rails with a model that uses STI, it's often easier to use one controller for all subclasses to control the download/access of the attached file. This interpolation allows you to reference that base class in the path/url.


jyurek:
This isn't really something that's applicable to a large amount of people, so I'd prefer to not have it default in Paperclip. Actually, things like this are exactly what [user-specified Interpolations](https://github.com/thoughtbot/paperclip/wiki/Interpolations) are for.

PR #2059, opened 2015-11-19
Update README with version requirement for aws-sdk

thiagogabriel:
The reference on README was to `master` while there was not release for paperclip 4.3.2.
Since it was just released, the reference was updated.


tute:
v4.3.2 is a different tree than `master`, which is now targeting v5. Thank you both for this commit and merge, but I'll revert it: paperclip v4.3.2 doesn't include AWS v2 support.

PR #2060, opened 2015-11-21
Fix condition to include Schema, using 'ActiveRecord::Base' instead o…

tonidas:
There are some gems that implements subclasses/submodules named ActiveRecord, that could wrongly trigger the Schema include.

Example: 
```ruby
module LogStasher
  module ActiveRecord
    class LogSubscriber < ::ActiveRecord::LogSubscriber
```
https://github.com/shadabahmed/logstasher/blob/7fbd7368734465a7bbd005d9fd2d7bb5c109d04e/lib/logstasher/active_record/log_subscriber.rb

To be safe that the real ActiveRecord is been used, I've added the ActiveRecord::Base class instead of the ActiveRecord class.

It's seems pretty safe in long term, since if they remove the Base subclass, all activerecord applications will be broken!

### DIFF
--- a/lib/paperclip/glue.rb
+++ b/lib/paperclip/glue.rb
@@ -8,7 +8,7 @@ module Paperclip
       base.extend ClassMethods
       base.send :include, Callbacks
       base.send :include, Validators
-      base.send :include, Schema if defined? ActiveRecord
+      base.send :include, Schema if defined? ActiveRecord::Base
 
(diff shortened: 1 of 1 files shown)


tute:
Thanks for your contribution! Have you run into this yourself?

tonidas:
Yeah. It took me sometime to figure out!

tute:
Thank you!

PR #2061, opened 2015-11-23
Modify the Gemfile for testing with mime-types 3

halostatue:
- The specs pass for the bits related to mime-types. I don’t use Paperclip
  otherwise, but there were a couple of local failures.
- Fix the appraisals, too.
- This will not fully work until mikel/mail is updated to support the same
  restriction, but I have PR open to expand mime-types support there, currently
  being reviewed for merge.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ gem 'pry'
 # Prevents bundler from taking a long-time to resolve
 group :development, :test do
   gem 'activerecord-import'
-  gem 'mime-types', '~> 1.16'
+  gem 'mime-types'
   gem 'builder'
   gem 'rubocop', require: false
 end
--- a/Appraisals
+++ b/Appraisals
@@ -1,9 +1,17 @@
 appraise "4.2.awsv2.1" do
   gem "rails", "~> 4.2.0"
   gem "aws-sdk", "~> 2.1.0"
+
+  group :development, :test do
+    gem 'mime-types', '>= 1.16', '< 4'
+  end
 end
 
 appraise "4.2.awsv2.0" do
   gem "rails", "~> 4.2.0"
   gem "aws-sdk", "~> 2.0.0"
+
+  group :development, :test do
+    gem 'mime-types', '>= 1.16', '< 4'
+  end
 end
--- a/gemfiles/4.2.awsv2.0.gemfile
+++ b/gemfiles/4.2.awsv2.0.gemfile
@@ -9,7 +9,7 @@ gem "aws-sdk", "~> 2.0.0"
 
 group :development, :test do
   gem "activerecord-import"
-  gem "mime-types", "~> 1.16"
+  gem "mime-types", ">= 1.16", "< 4"
   gem "builder"
   gem "rubocop", :require => false
 end
--- a/gemfiles/4.2.awsv2.1.gemfile
+++ b/gemfiles/4.2.awsv2.1.gemfile
@@ -9,7 +9,7 @@ gem "aws-sdk", "~> 2.1.0"
 
 group :development, :test do
   gem "activerecord-import"
-  gem "mime-types", "~> 1.16"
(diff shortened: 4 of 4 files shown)


tute:
What were the local failures? What issue is this solving? Thanks for your contribution!

halostatue:
1. I don’t remember the local failures. They were probably related to the fact that I made the changes to allow for mime-types 3 and only ran `rake` at that point. There may be set-up that I don’t know of. This is passing on your CI, so it doesn’t much matter.

2. This solves the issue that paperclip is only allowing mime-types 1 and mime-types 2, when mime-types 3 was released at the end of November. mime-types 2 will only see data and security updates only until November 2017. mime-types 3 is where future development will be.

Gemfile
```diff
@@ -9,7 +9,7 @@ gem 'pry'
 # Prevents bundler from taking a long-time to resolve
 group :development, :test do
   gem 'activerecord-import'
-  gem 'mime-types', '~> 1.16'
+  gem 'mime-types', '>= 1.16', '< 4'
```
tute: Can these be simplified to '~> 3.0'?

halostatue: Probably not. mime-types 3.x is Ruby 2.0+ only. mime-types 2.x is 1.9+ only. mime-types 1.x supports pretty much any version of Ruby, but is no longer receiving any updates at all (it hit EOL on 27 October 2015). When Rails 5 is released, your appraisal gemfile can absolutely restrict to `~> 3.0` because Rails 5 will only support Ruby 2.2+.

tute: Great! Can you rebase on top of master adding this comment to the commit message? Thanks so much!

halostatue: Done with a slightly different change. I have copied the `mime-types` restriction to the Appraisal so that, as you prefer, you can start varying these for each version of Rails + AWS supported by Paperclip.

tute:
> mime-types 1.x supports any version of Ruby, but no longer receives any updates (it hit EOL on 27 October 2015).

In that case, we should remove it as a dependency. Can that be done safely? If not, what backwards incompatible change would that introduce?

halostatue:
The only external API change between mime-types 1.x and 2.x is that Ruby 1.8.7 was dropped from support. Mostly the same between mime-types 2.x and 3.x (dropping Ruby 1.9 support, but also dropping methods deprecated in 2.x). Paperclip uses one of the most stable APIs that will ever exist on mime-types (if I break that one, I break the *raison d’être* of mime-types).

tute:
> The bigger question is whether or not downstream gems, projects, or people who use Paperclip will have other things which restrict to mime-types 1.x (I’m trying to remove it permanently…). That I don’t have an answer for at the moment.

I don't think this is important. If they do, they should stick with paperclip versions which allow for that --equally old, potentially insecure.

Newer versions of paperclip (master, now pointing to v5) should drop support of outdated dependencies.

tute:
BTW, thanks for taking on that project! :) :tada:

maclover7:
Please rebase.

halostatue:
Well, that makes the changes simpler. ;)

maclover7:
:+1: thanks for your contribution!

maclover7:
@halostatue Do you want to write a PR that restricts the `mime-types` dependency in the gemspec to `~> 3`?

halostatue:
@maclover7 mikel/mail that allows for mime-types 3 has not yet been released as anything more than a RC, so you could find a situation where people have conflicting dependencies that can’t be resolved due to this sort of PR. IMO, this is something that you guys can add for whatever version will be released that supports Rails 5 only. (I am also providing security/data updates to mime-types 2 for another 21 months or so, so while only new features will show up with mime-types 3, mime-types 2 is not dead *yet*.)

maclover7:
:+1: thanks for the info @halostatue!

PR #2064, opened 2015-11-25
upgrading aws-sdk to v2 requires changing 'public_read' to 'public-read'

betesh:
See https://github.com/thoughtbot/paperclip/issues/1974


betesh:
I edited in the browser, so it came out as 2 separate PRs but I was able to merge one into the other.  Funny how that worked out.

tute:
Thank you! Merged in. I thought the symbol has to be a string, too.

betesh:
It can be a Symbol or String in either version, I think.  If I implied otherwise, that was a typo.

PR #2065, opened 2015-11-30
fix S3 adapter runtime constant definition not being thread-safe

erkki:
After debugging seemingly random exceptions on DEFAULT_PERMISSION not being defined in a threaded downloader, it became clear the S3 adapter is not thread safe.

Here is a sample run with 2 threads and debug prints:
```
#<Thread:0x007f925d34eeb8> before defined? check AWS_CLASS
#<Thread:0x007f925d34eeb8> before const_set AWS_CLASS
#<Thread:0x007f925d34eeb8> after const_set AWS_CLASS
#<Thread:0x007f925d34ed78> before defined? check AWS_CLASS
#<Thread:0x007f925d34ed78> before usage of const DEFAULT_PERMISSION
#<Thread:0x007f925d34eeb8> before const_set DEFAULT_PERMISSION
#<Thread:0x007f925d34eeb8> after const_set DEFAULT_PERMISSION
#<Thread:0x007f925d34eeb8> before usage of const DEFAULT_PERMISSION
#<Thread:0x007f925d34eeb8> after usage of const DEFAULT_PERMISSION
```
The added Thread#exclusive block is probably too wide and blunt of an fix but should be fine until a better one comes along.

Decided to use 10 threads in the spec after making sure of a good-enough failure rate (it is still possible to get false-positives but they should be rare).

### DIFF
--- a/lib/paperclip/storage/s3.rb
+++ b/lib/paperclip/storage/s3.rb
@@ -113,21 +113,23 @@ module Paperclip
 
     module S3
       def self.extended base
-        unless defined?(AWS_CLASS)
-          begin
-            require 'aws-sdk'
-            const_set('AWS_CLASS', defined?(::Aws) ? ::Aws : ::AWS)
-            const_set('AWS_BASE_ERROR',
-              defined?(::Aws) ? Aws::Errors::ServiceError : AWS::Errors::Base)
-            const_set('DEFAULT_PERMISSION',
-              defined?(::AWS) ? :public_read : :'public-read')
+        Thread.exclusive do
+          unless defined?(AWS_CLASS)
+            begin
+              require 'aws-sdk'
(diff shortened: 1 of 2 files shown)


erkki:
Any feedback here? I'm looking at the few travis failures, at least one  seems non-related.

tute:
Hey, thank you! Is this still relevant after recent changes to this file? If so, can you please rebase your changes? Thank you!

erkki:
Sure thing, let me check asap and get back to you, thanks!

tute:
This code has been removed since then. Thank you for your time and sharing your work! ❤️

erkki:
Sorry about the delay in response, yes, confirmed that this is not an issue with the code removed. Thanks for handling :)

Issue #2049, opened 2015-11-12
V5.0 Release Plan


maclover7:
There will be several breaking changes in Paperclip 5.0.0, so it's important that proper deprecation cycles are followed and good documentation is released to make the upgrade cycle as easy as possible.

**TODO:**
- [ ] Release a minor version FIRST (4.3.x) with the following changes:
  - [x] Deprecation warning if the user is still on aws-sdk v1, say v1 support will be dropped in v5.0.0
  - [x] Deprecation warning if the user is still on Rails 3.2.x, say 3.2.x support will be dropped in v5.0.0
- [x] Release a major version SECOND (5.0.0)
  - [x] Actual deprecation of aws-sdk v1 (this is a breaking change) (#2075)
  - [x] Actual deprecation of Rails 3.2 and 4.1 (this is a breaking change) (#2074)
  - [x] Remove locales from main repository (#1996)
  - [x] Add support for Rails 5.0.0 (#1976)
  - [x] Update documentation (#1976)

Deprecation notices should try and follow this style:
```ruby
require 'active_support/deprecation'
ActiveSupport::Deprecation.warn "deprecation notice goes here!"
```

As far as testing deprecation notices, here is a code example for that too:
```ruby
require 'active_support/testing/deprecation'

describe CoolClass do
  include ActiveSupport::Testing::Deprecation

  assert_deprecated do
    # code that will trigger deprecation goes here
  end
end
```

#1991 should be finished off whenever next 4.3.x release has been shipped.

ping @tute and @jyurek

tute:
:+1: :clap: :heart:

connorshea:
Now that Rails 5.0.0.beta1 is out, any updates on Paperclip 5.0.0?

tute:
> Now that Rails 5.0.0.beta1 is out, any updates on Paperclip 5.0.0?

We are working on the checlist above. If there is no related issue, feel free to grab an item and move forward with it! I'd like to have a release in ~2 weeks or less.

tute:
@maclover7 I forgot: Rails 5 has to work on the next 4.x release of paperclip too. Otherwise people have to upgrade in tandem both paperclip _and_ Rails. Will add it to the checklist.

tute:
Notes from an email asking questions on how to implement deprecations:

> Where exactly you want the warning? I’ve not found a correct “method” to be the one that asserts we have a warning. Can you point me to a place in the code where deprecation should be thrown

[...]

PR #2074, opened 2015-12-22
Removes Rails 3.2 and 4.1 support

tute:
In preparation for Rails 5, which, when it is released, will make Rails
3.2 and 4.1 unsupported.

See:

* http://weblog.rubyonrails.org/2015/12/18/Rails-5-0-beta1/
* https://github.com/thoughtbot/paperclip/issues/2049


maclover7:
@tute - We're going to have to revert this for right now. We need to get deprecations into a 4.x point release before we go ahead with any breaking changes :)

maclover7:
Wait -- never mind. Didn't realize master is already backwards incompatible, and that we are using the `v4.3` branch for 4.3.x releases. Please disregard :)

tute:
v4 is its own branch, and lives behind master. This is fine, the 4.2.x cut will be done with the warnings.

tute:
Right. :+1:

maclover7:
:+1:

PR #2075, opened 2015-12-22
Remove AWS v1 support

tute:
TODO: Simpify lib/paperclip/storage/s3.rb

See:

* https://github.com/thoughtbot/paperclip/issues/2049


maclover7:
As soon as CI passes, :+1:

tute:
@maclover7 AWS v1 was a dependency hell (ask @jyurek :) ), not a code maintenance hell. I thought that S3 class was way more complex but the diff is simple. I'm happy, anyway, with simplifying dependencies; but if we leave too many people out with this I'm not against reintroducing it back.

Now it has a green CI, and I'll merge. Thanks!

PR #2082, opened 2016-01-02
Remove unused vars `column_(type|options)` from `remove_attachment` method

keram:
These variables were introduced in:
https://github.com/thoughtbot/paperclip/commit/6ea7c26512434a
but their usage were canceled in:
https://github.com/thoughtbot/paperclip/commit/c740fb171fe


maclover7:
:+1: thanks!

PR #2083, opened 2016-01-02
Replace File.exists? with File.exist? in paperclip.gemspec

keram:
File.exists? is deprecated and produces unecessary warning.


maclover7:
Thanks!

Issue #2040, opened 2015-11-07
Double reporting validation errors


iDiogenes:
I am running rails 4.1 with the head of paperclip master and the size validation errors are displaying two of the same error. I have tried validates_attachment and validates_attachment_size and both produce the same issue:

```
validates_attachment_size :logo, less_than: 1.megabytes
validates_attachment :logo, content_type: { content_type: /\Aimage/ }, size: { less_than: 1.megabytes }
```

![screen shot 2015-11-06 at 6 51 24 pm](https://cloud.githubusercontent.com/assets/45213/11013245/8dd243dc-84b9-11e5-8b69-f0db88c7d7cf.png)

evgenii:
Yes, guys this is the problem.
Why are you doing that: https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/validators/attachment_size_validator.rb#L29

Currently I've add after_filter in my model for remove duplicates, but this is not a good approach

tute:
Closing as duplicate of https://github.com/thoughtbot/paperclip/issues/1367. Thank you!

Issue #1367, opened 2013-10-25
Making duplicate error messages configurable


scottsherwood:
Hi,

After a little investigation into why validation messages are being duplicated, paperclip now sets errors on both the base and attribute. #1253 

I'm suggesting that we make this configurable so that those of us who only want them on base (or the attr) can easily allow this. As suggested ( @djcp  ) I'm opening an issue so that this can be discussed and get everyone's feedback.

So any suggestions on how this could be implemented? What the defaults should be? If its even a good idea to add the config? etc..

maclover7:
Hi @scottsherwood ! Is this still an issue for you in Paperclip; I know this issue is from approximately 2 years ago. If it is still an issue, would you be willing to send in a PR to solve the problem? Thanks!

scottsherwood:
Hi @maclover7

Yes this is still an issue with paperclip, but I'm afraid my workload at the minute is a little to high and i'm unlikely to have any time to work on this anytime soon.

Sorry
Scott

tute:
Will close this issue in favor of discussions in the two open PRs. Thank you all.

tute:
Reopening this issue, as the two PRs were not good solutions for the problem. The issue persists, and a suggested workaround in application code is:

```ruby
class Attachment < ActiveRecord::Base
  after_validation :clean_up_paperclip_errors

  private

  # This is because paperclip duplicates error messages
  # See: https://github.com/thoughtbot/paperclip/pull/1554 and
  # https://github.com/thoughtbot/paperclip/commit/2aeb491fa79df886a39c35911603fad053a201c0
  def clean_up_paperclip_errors
    errors.delete(:file)
  end
end
```

txapelgorri:
@tute: thanks for this workaround, works great :)

brendon:
Hi @tute, I think a proper configurable fix should be pursued because the solution you mention here: https://github.com/thoughtbot/paperclip/issues/1367#issuecomment-217934878 fails to take into account the `presence` validation that is only added to `base`, thus this error gets deleted and we end up with a similar problem that form builder users were having with regards to missing or hidden messages.

I think we all agree that this should be something handled internally to Paperclip rather than scattering these callbacks across a bunch of models in an app, especially considering the workaround isn't 100% working.

Perhaps we could decide on what the configuration option looks like, then work from there on a PR?

Something like:?

```ruby
Paperclip.options[:add_validation_errors_to] = :base
[...]

PR #2090, opened 2016-01-15
Add global and per validation options for duplicating errors on base …

aouji:
…attribute.

Adds an option to turn duplication of error messages on the base attribute on/off globally or per validation. The option is on globally by default.


aouji:
This should fix #2040 and #1367 .

aouji:
Let me know if the Hound has to be convinced 100%.

aouji:
@tute Can you please let me know if I am taking the right approach here? Just wondering if this is in line with what thoughtbot would want and if I should do another round of addressing the houndci-bot comments.

BTW, I think the `errors.delete(:file)` approach mentioned in https://github.com/thoughtbot/paperclip/pull/1554 has the side effect of removing errors that are not duplicated (e.g. errors from the presence and spoof detection validations).

tute:
@aouji sorry I didn't get to it this week. Hopefully next one! Thank you very much for your work.

aouji:
Sorry to ping you again @tute . Have you had time to take a look at this?

tute:
Thanks for pinging @aouji. Not yet, hopefully next week.

tute:
@aouji I just closed https://github.com/thoughtbot/paperclip/pull/1554#issuecomment-200666161 with that linked comment that states a solution for client applications that need to override this behavior, and my rationale on why I think that PR (solving this issue) should not be merged into paperclip.

Your PR adds 300 lines of code, and the ability for users to turn off the duplication of error messages onto base. It doesn't take into account the users who actually want to have the errors on base, but not on each field. So it adds complexity to the project, but it doesn't solve the underlying issue (or it does, for one of the three different use cases).

I appreciate your time and effort invested in triaging this issue, and proposing and sharing a fix with us. But I don't think this PR is a good alternative solution for current paperclip behavior, and thus I will close it as well.

As stated in the other PR, I'll consider a fix that, without adding much complexity, takes care of both use cases.

Thank you very much, again. ❤️

aouji:
Thanks for the detailed explanation @tute! 

Regarding the ~300 lines of change in the pull, ~30 lines are code changes and the rest are tests. That being said, you are completely right about the use cases. My code does not cover users wanting to keep base but remove the other ones. I might make another try at covering all scenarios if I find the time.

Thanks for maintaining paperclip. Without it a lot of rails apps - including mine - will be in big trouble!

Issue #357, opened 2010-11-16
Don't fingerprint when there is no fingerprint column


iain:
While investigating the issue regarding Rails >= 3.0.2, I found that paperclip always calculates the fingerprint, even though there is no fingerprint column to store it in. Should it still be calculated then? Or is it used for some other method? Otherwise, it just takes up time.

PR #2091, opened 2016-01-15
Skip calculating fingerprint when it is not used

mattbrictson:
The MD5 sum of an attachment is only necessary if the model has a corresponding `_fingerprint` column. If this column is absent, there is no need to calculate the MD5, which can be an expensive operation for large files.

Accomplish this by deferring the fingerprint calculation using a block. If the `_fingerprint` column is absent, the block is never called, and the calculation is avoided.

This issue was fixed once before (see #357, #696), but a subsequent refactor seems to have reintroduced the problem.

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -427,7 +427,7 @@ module Paperclip
     def assign_attributes
       @queued_for_write[:original] = @file
       assign_file_information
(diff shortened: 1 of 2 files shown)


maclover7:
:+1: Looks good to me - thanks for your contribution!

maclover7:

PR #2093, opened 2016-01-22
Bugfix: Dynamic fog directory

### LINKED ISSUE #2018: Fog storage doesn't consistently check for dynamic :fog_directory
`Paperclip::Storage::Fog#host_name_for_directory` assumes a String-like object is set and doesn't check if `@options[:fog_directory]` is callable, while `Paperclip::Storage::Fog#directory` does. I would recommend extracting a new method with the condition and refactoring the other two methods to call it.

See https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/storage/fog.rb#L191 and https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/storage/fog.rb#L220

ravinggenius:
`Paperclip::Storage::Fog#host_name_for_directory` assumes a String-like object is set and doesn't check if `@options[:fog_directory]` is callable, while `Paperclip::Storage::Fog#directory` does. I would recommend extracting a new method with the condition and refactoring the other two methods to call it.

See https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/storage/fog.rb#L191 and https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/storage/fog.rb#L220

Fixes #2018.

### DIFF
--- a/lib/paperclip/storage/fog.rb
+++ b/lib/paperclip/storage/fog.rb
@@ -195,10 +195,10 @@ module Paperclip
       end
 
       def host_name_for_directory
-        if @options[:fog_directory].to_s =~ Fog::AWS_BUCKET_SUBDOMAIN_RESTRICTON_REGEX
-          "#{@options[:fog_directory]}.s3.amazonaws.com"
+        if directory_name.to_s =~ Fog::AWS_BUCKET_SUBDOMAIN_RESTRICTON_REGEX
+          "#{directory_name}.s3.amazonaws.com"
         else
-          "s3.amazonaws.com/#{@options[:fog_directory]}"
+          "s3.amazonaws.com/#{directory_name}"
         end
       end
 
@@ -224,13 +224,15 @@ module Paperclip
       end
 
       def directory
-        dir = if @options[:fog_directory].respond_to?(:call)
+        @directory ||= connection.directories.new(key: directory_name)
+      end
+
+      def directory_name
+        if @options[:fog_directory].respond_to?(:call)
           @options[:fog_directory].call(self)
         else
           @options[:fog_directory]
         end
-
-        @directory ||= connection.directories.new(:key => dir)
       end
 
       def scheme
--- a/spec/paperclip/storage/fog_spec.rb
+++ b/spec/paperclip/storage/fog_spec.rb
@@ -418,6 +418,9 @@ describe Paperclip::Storage::Fog do
           assert @connection.directories.get(@dynamic_fog_directory).inspect
         end
 
+        it "provides an url using dynamic bucket name" do
+          assert_match(/^https:\/\/dynamicpaperclip.s3.amazonaws.com\/avatars\/5k.png\?\d*$/, @dummy.avatar.url)
+        end
       end
 
       context "with a proc for the fog_host evaluating a model method" do
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,5 @@
+* Bugfix: Dynamic fog directory option is now respected
+
 5.0.0.beta1
 
 * Drop support to end-of-life'd ruby 2.0.


ravinggenius:
Feel free to cherry-pick around the first two commits, but I find those to be extremely useful.

spec/paperclip/storage/fog_spec.rb
```diff
@@ -418,6 +418,9 @@ def custom_method
           assert @connection.directories.get(@dynamic_fog_directory).inspect
         end
 
+        it "provides an url using dynamic bucket name" do
+          assert_match(/^https:\/\/dynamicpaperclip.s3.amazonaws.com\/avatars\/5k.png\?\d*$/, @dummy.avatar.url)
```
ravinggenius: No it isn't.

lib/paperclip/storage/fog.rb
```diff
@@ -217,14 +217,16 @@ def connection
...
+        if @options[:fog_directory].respond_to?(:call)
           @options[:fog_directory].call(self)
         else
           @options[:fog_directory]
         end
+      end
 
-        @directory ||= connection.directories.new(:key => dir)
+      def directory
+        @directory ||= connection.directories.new(:key => directory_name)
```
ravinggenius: I would, but this isn't new code.

spec/paperclip/storage/fog_spec.rb
ravinggenius: Sorry my last comment was a bit sarcastic. I'm leaving this as is for now as it matches the established code style.

.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{rb,gemspec}]
+indent_size = 2
+indent_style = space
+
+[Gemfile]
+indent_size = 2
+indent_style = space
```
tute: We don't need this file.

lib/paperclip/storage/fog.rb
tute: Let's use new hash syntax here.

lib/paperclip/storage/fog.rb
```diff
@@ -217,14 +217,16 @@ def connection
         @connection ||= ::Fog::Storage.new(fog_credentials)
       end
 
-      def directory
-        dir = if @options[:fog_directory].respond_to?(:call)
+      def directory_name
```
tute: I'd define this after `directory`, as it's used by it and file will read better.

spec/spec_helper.rb
```diff
@@ -37,6 +37,9 @@
   config.include ModelReconstruction
   config.include TestData
   config.extend VersionHelper
+  config.filter_run focus: true
+  config.filter_run_excluding blur: true
+  config.run_all_when_everything_filtered = true
```
tute: Do we need these changes? I'd discuss them in another PR.

tute:
Left some comments, looking good. Thank you!

spec/spec_helper.rb
ravinggenius: We don't need this change for this bugfix; I just like to run a single test without screwing around on the command line looking for line numbers and typing out file names. I've removed the commit.

ravinggenius:
Thanks for looking into this. I rebased back onto master, made the requested changes and force-pushed my branch.

tute:
Thank you! Can you please add your PR description as commit message (adjusted: you don't recommend, you implement :) ), and edit the NEWS file? Aand we are ready! Thanks for catching this.

ravinggenius:
I rebase onto master, verified everything still works, and force-pushed again. The `NEWS` file is up-to-date as well. Am I missing anything else?

tute:
Thank you! :)

PR #2094, opened 2016-01-26
Remove attr_protected related specs (was removed in rails 4)

kevcha:
Since support for Rails 3.2 was removed (https://github.com/thoughtbot/paperclip/pull/2074), those specs are not more needed

### DIFF
--- a/spec/support/rails_helpers.rb
+++ b/spec/support/rails_helpers.rb
@@ -1,7 +1,0 @@
-module RailsHelpers
-  module ClassMethods
-    def using_protected_attributes?
-      ActiveRecord::VERSION::MAJOR < 4
-    end
-  end
-end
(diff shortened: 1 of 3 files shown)


spec/support/rails_helpers.rb
```diff
@@ -1,7 +1,4 @@
 module RailsHelpers
   module ClassMethods
-    def using_protected_attributes?
-      ActiveRecord::VERSION::MAJOR < 4
-    end
   end
 end
```
tute: Do you think we should delete this file entirely?

tute:
Thank you! Left one question, ready for merge after we address it.

spec/support/rails_helpers.rb
kevcha: I think yes, it should be rewrite in future if needed

maclover7:
:+1: Thank you for your contribution!

maclover7:

Issue #2098, opened 2016-01-31
Problem when using paperclip without active_record


kt0:
Latest release ([4.3.3](https://github.com/thoughtbot/paperclip/releases/tag/v4.3.3)) is problematic when using paperclip with mongoid, that there is no active_record.

This [single commit](https://github.com/thoughtbot/paperclip/commit/2e6335532ac24012cf473b60994d09f14952b891) is checking for deprecations and relies on `ActiveRecord` constant being avaible.

Version prior to this (4.3.2) is working as expected.

betesh:
The quick fix would be to use ActiveModel or ActiveSupport (which are paperclip dependencies) instead of ActiveRecord.

kevcha:
Nice catch @kt0
As @betesh says, relying on ActiveModel or ActiveSupport versions instead of ActiveRecord would fix it. I'm gonna make a pull request for this

maclover7:
Closed via #2100. Version `4.3.4` has been released to solve this.

PR #2100, opened 2016-02-02
Check ActiveModel instead of ActiveRecord for deprecations

kevcha:
As @kt0 noticed in #2098, my commit https://github.com/thoughtbot/paperclip/commit/2e6335532ac24012cf473b60994d09f14952b891 broke  compatibility for non ActiveRecord user.
Checking ActiveModel version instead of ActiveRecord fixes this since it's a paperclip dependency


maclover7:
:+1: Thanks for submitting this! Sorry for the backwards compatible issues here.

PR #2103, opened 2016-02-04
Allow Paperclip.run to recieve a page number as an optional argument

rpbaptist:
This allows for specification of which page to convert when thevattachment is a multi-page document.

With Imagemagick, running `convert[1] document.pdf document.png` allows the generated thumbnail to be from the second page instead of the first which it defaults to. I implemented this by overriding the `Processor.convert` method, but thought I might as well submit a PR.

It is also possible to supply multiple pages in the format of `convert[1,3,4]` and I considered implementing this, but I don't think Paperclip expects multiple thumbnails on one file and it doesn't fit my own use case.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,6 @@
 master:
 
+* Improvement: Add `read_timeout` configuration for URI Adapter download_content method.
 * README adjustments for Ruby beginners (add links, elucidate model in Quick Start)
 * Bugfix: Now it's possible to save images from URLs with special characters [#1932]
 * Fix a nil error in content type validation matcher [#1910]
--- a/lib/paperclip.rb
+++ b/lib/paperclip.rb
@@ -97,7 +97,8 @@ module Paperclip
       :log_command => true,
       :swallow_stderr => true,
       :content_type_mappings => {},
-      :use_exif_orientation => true
+      :use_exif_orientation => true,
+      :read_timeout => nil
     }
   end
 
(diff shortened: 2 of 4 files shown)


rpbaptist:
@houndci-bot and I are having a great conversation on the merits of this PR.

jyurek:
I don't understand this. You're adding those to the binary invocation itself rather than the arguments? That doesn't seem right to me.

rpbaptist:
Thank you for responding.

I agree it would have been nice as an argument. As far as I have seen, Imagemagick expects the format to be `convert[1]` instead of `convert` when giving a page number.

The other option would be to change it in [`Paperclip::Helpers.run`](https://github.com/thoughtbot/paperclip/blob/106a0257f4b3c647791518bc499054d9f2492f43/lib/paperclip/helpers.rb#L31-L31).

```
page_number = local_options.delete(:page_number)
page_number ? command = "#{cmd}[#{page_number}]" : cmd
Cocaine::CommandLine.new(command, arguments, local_options).run(interpolation_values)
```

Either way, the command itself needs to be changed and since it isn't a valid option, it shouldn't be send to the command line along with `local_options`.

I don't see a nicer way of doing this until Imagemagick supports the page number as a parameter.

bdewater:
The official way is to add the page number to the file name, not the convert command. E.g.

    convert document.pdf[1] document.png

 See the "Selecting Frames" section on http://www.imagemagick.org/script/command-line-processing.php

rpbaptist:
@bdewater Thanks, I will close this and submit a new PR.

PR #2108, opened 2016-02-08
Remove deprecations

maclover7:
This will be released in v4.3.5.


maclover7:
Travis is green -- merging!

sikachu:
@maclover7 thanks!

tute:
This should have been released in v5, not 4.3, right? To help people upgrade (think rspec 2.99).

tute:
Yes, people running on latest paperclip don't get the deprecation warnings necessary for upgrading to v5. I think we need to revert this commit and release another patch release.

sikachu:
@tute it should be done after `v5.0.0` is out, otherwise user will just keep getting message nagging them. 5.0 and 4.3.6 with deprecation message can go out at the same time.

tute:
I see. I was inferring people will upgrade. What do you think of a 4.99 version, with only the deprecation warnings, à la rspec, so we explicitly state that it's the latest version when you want to move on to 5?

Issue #1665, opened 2014-10-01
Incorrect values in attachment_content_type and attachment_file_name db columns


sevab:
I have a photo attachment with several styles all of which convert the image to a jpg, e.g.:
```ruby
styles: { 
:original, { geometry: "1500x1500>", format: :jpg },
:large, { geometry: "1000x1000>", format: :jpg },
:medium, { geometry: "650x650#", processors: [:cropper], format: :jpg },
}
```

The model also has a `before_post_process` that renames photo's filename to a standard name:
```ruby
def rename_photo
    extension = File.extname(photo_file_name).downcase
    self.photo.instance_write :file_name, "original#{extension}"
end
```
With this set up whatever file I upload, be it a JPEG or a PNG, it gets correctly converted to a JPEG: a JPEG is uploaded to S3 and attachment's `url` method returns a url ending with `.jpg`. However, the information stored in the `attachment_file_name` and `attachment_content_type` columns in db contain incorrect information for formats other than JPG. Instead of `original.jpg` and `image/jpeg` they incorrectly contain `original.png` and `image/png`.

tute:
Thanks for your report @sevab. Is this still an issue with paperclip? How can we replicate it?

tute:
Closing due to lack of activity.

garnieretienne:
@tute , Same issue here. 
I build a sample app showing the bug:
https://github.com/garnieretienne/paperclip_bug

maclover7:
Can you try and do a source dive into Paperclip to try and identify the source of this bug?

garnieretienne:
_I'm not familiar with the paperclip source_

@maclover7 My guess was to patch this [part of code](https://github.com/thoughtbot/paperclip/blob/bd016009dbf74fc8f999e78a68c9e5869eb0dd6a/lib/paperclip/attachment.rb#L467):

```ruby
# lib/paperclip/attachment.rb#466

def reset_file_if_original_reprocessed

      # That only update the file size after the original file has been processed
      # instance_write(:file_size, @queued_for_write[:original].size)
      # Replaced by that to update the `file_name`, `content_type`, `file_size`
      assign_file_information

      assign_fingerprint { @queued_for_write[:original].fingerprint }
      reset_updater
    end
```

To test if my change was enough I was looking to find an existing test for the `reset_file_if_original_reprocessed` method and only found this:

```ruby
context "An attachment with a default style and an extension interpolation" do
    before do
[...]

PR #2117, opened 2016-02-21
Fix incorect values for `attachment_content_type` and `attachment_file_name` in some case

garnieretienne:
Fix the 'file_name' and 'mime_type' attributes for attachments with an extension interpolation on the original style.

See issue https://github.com/thoughtbot/paperclip/issues/1665.


tute:
Thanks so much for your help with this, @garnieretienne. CI is failing, can you please fix it, and rebase your branch on top of latest master? Thank you! 😃

PR #2127, opened 2016-03-09
Update example to add mime-types to file extensions

### LINKED ISSUE #1737: update doc: MIME::Types#index_extensions is deprecated and will be private.
The warning shows up when I copy it from document. 
But I don't find method to avoid it

mime-types (2.4.3)

bdewater:
Fixes #1737.

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,14 +445,13 @@ will not cause errors to be raised.
 
 This can sometimes cause false validation errors in applications that use custom
 file extensions. In these cases you may wish to add your custom extension to the
-list of file extensions allowed for your MIME type configured by the `mime-types`
-gem:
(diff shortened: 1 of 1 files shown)


tute:
Thanks! Since when do we need this doc change? Can you please add context to your commit message, with a link to the change? Then we'll merge. Thank you!

bdewater:
I'm not sure since when that gem started giving those warnings - the old method still works on version <3 but with master now targeting >= 1.16 && < 4 it seemed like time to move away from deprecated functionality. I'll update the commit message in a couple of hours.

bdewater:
Updated :)!

tute:
Thank you!

PR #2126, opened 2016-03-09
Backport fix for excessive logging from media type spoof detector 

bdewater:
Backport 8339e0f (#2017) to 4.3 to prevent excessive logging, which [might be abused](https://cwe.mitre.org/data/definitions/779.html). Fixes #2125. 

Please release 4.3.6 as soon as possible to prevent abuse.

PR #2131, opened 2016-03-12
Add test to prevent regression regarding excessive logging

bdewater:
This is the test from #2126 for the master branch.

### DIFF
--- a/spec/paperclip/media_type_spoof_detector_spec.rb
+++ b/spec/paperclip/media_type_spoof_detector_spec.rb
@@ -44,9 +44,18 @@ describe Paperclip::MediaTypeSpoofDetector do
     end
   end
 
-  it "rejects a file if named .html and is as HTML, but we're told JPG" do
-    file = File.open(fixture_file("empty.html"))
-    assert Paperclip::MediaTypeSpoofDetector.using(file, "empty.html", "image/jpg").spoofed?
+  context "file named .html and is as HTML, but we're told JPG" do
+    let(:file) { File.open(fixture_file("empty.html")) }
+    let(:spoofed?) { Paperclip::MediaTypeSpoofDetector.using(file, "empty.html", "image/jpg").spoofed? }
+
+    it "rejects the file" do
+      assert spoofed?
+    end
+
+    it "logs info about the detected spoof" do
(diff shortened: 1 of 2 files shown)


spec/paperclip/media_type_spoof_detector_spec.rb
```diff
@@ -44,9 +44,18 @@
...
+  context "file named .html and is as HTML, but we're told JPG" do
+    let(:file) { File.open(fixture_file("empty.html")) }
+    let(:spoofed?) { Paperclip::MediaTypeSpoofDetector.using(file, "empty.html", "image/jpg").spoofed? }
+
+    it "rejects the file" do
+      assert spoofed?
+    end
+
+    it "logs info about the detected spoof" do
+      Paperclip.expects(:log).with('Content Type Spoof: Filename empty.html (image/jpg from Headers, ["text/html"] from Extension), content type discovered from file command: text/html. See documentation to allow this combination.')
```
tute: I think capture will read better. Also, by matching what we want to match we'll make the spec less brittle to log string changes. Do you think this would improve the spec?

```ruby
output = capture(:stdout) do
  run_arbitrary_code
end
assert output =~ /pattern/
```

Thank you!

bdewater: I agree that capturing std_out is better, testing behaviour instead of implementation. The brittleness might not be so bad in this case. A pattern could match against those very long lines of mime-types data which is the regression I am trying to prevent with this test. I'll try and figure something out in the weekend.

tute:
@bdewater do you have availability to rebase and get this PR ready? Thank you! :)

bdewater:
@tute I've rebased my original PR. I've tried your suggestion and the following to capture log output:
```ruby
it "logs info about the detected spoof" do
  aggregate_failures do
    expect { spoofed? }.to output(/image\/jpg from Headers/).to_stdout
    expect { spoofed? }.to output(/\["text\/html"\] from Extension/).to_stdout
    expect { spoofed? }.to output(/file command: text\/html/).to_stdout
  end
end
```
but the output is always an empty string. I don't have time to work on this for at least the next two weeks unfortunately.

tute:
Thank you! :)

PR #2143, opened 2016-03-23
close+unlink leaking tempfiles + add specs to prevent regression

erkki:
This fixes at least some of lingering issues reported in #1326 ; I'll need to run this through some testing in production to see if there's any breakages outside of what the specs cover.

In the meanwhile would appreciate if anyone has any comments or suggestions.

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -523,6 +523,10 @@ module Paperclip
         @queued_for_write[name] = style.processors.inject(@queued_for_write[:original]) do |file, processor|
           file = Paperclip.processor(processor).make(file, style.processor_options, self)
           intermediate_files << file unless file == @queued_for_write[:original]
+          # if we're processing the original, close + unlink the source tempfile
+          if name == :original
+            @queued_for_write[:original].close(true)
+          end
           file
         end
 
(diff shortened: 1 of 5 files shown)


tute:
@erkki did you run tests in production? How did your branch behave?

If you'd like to move forward with this, let's please rebase the branch on top of latest master, and adhere to the style guides (Hound) whenever possible. Thank you! 😃

erkki:
We've been running it in production ever since, seems to work for us. Let me take a look again and rebase.

erkki:
There's new regressions on latest master so it'll take me some time to debug, stay tuned :)

mike-burns:
Hi @erkki , I didn't follow the code closely enough to know what the next step should be, but this test

```sh
rspec spec/paperclip/integration_spec.rb:126
```

only calls `post_process_style` with `:thumb_small` or `:thumb_large`. Not sure why it's failing or what it was supposed to do, sorry.

erkki:
Sorry, this is ancient history at this point, I don't remember the specifics anymore. The regression test probably serves as a good anchor (ie files should not leak, period).

mike-burns:
Note to myself (I've been trying to figure this out on Fridays):

`#reprocess!` calls `#assign` which calls `Paperclip.io_adapters.for` which creates the tempfile. `#assign` then calls `#post_process_file` which (somehow?) calls `MediaTypeSpoofDetectionValidator#validate_each` which calls `Paperclip.io_adapters.for` which creates another tempfile. That first tempfile is the one that leaks.

PR #2146, opened 2016-03-24
Removes dead code

tute:
cc @kitop in case you spot more pieces.


tute:
Will leave this open for a few days until we remove the many pieces we don't need anymore.

PR #2155, opened 2016-03-29
allow user to specify frame_index for generating thumbnails from videos or pdfs

jacobbullock:
add optional parameter in style options to specify a frame_index to use for generating thumbnails from videos or pdfs.  If no value is set, it will default to 0 to maintain default paperclip functionality.

### DIFF
--- a/lib/paperclip/thumbnail.rb
+++ b/lib/paperclip/thumbnail.rb
@@ -77,7 +77,11 @@ module Paperclip
 
         parameters = parameters.flatten.compact.join(" ").strip.squeeze(" ")
 
-        success = convert(parameters, :source => "#{File.expand_path(src.path)}#{'[' + @frame_index.to_s + ']' unless animated?}", :dest => File.expand_path(dst.path))
+        success = convert(parameters, 
+          :source => "#{File.expand_path(src.path)}#{'[' + @frame_index.to_s + ']' unless animated?}", 
+          :dest => File.expand_path(dst.path)
+          )
+          
       rescue Cocaine::ExitStatusError => e
         raise Paperclip::Error, "There was an error processing the thumbnail for #{@basename}" if @whiny
       rescue Cocaine::CommandNotFoundError => e
--- a/lib/paperclip/thumbnail.rb
+++ b/lib/paperclip/thumbnail.rb
@@ -42,7 +42,7 @@ module Paperclip
       if @auto_orient && @current_geometry.respond_to?(:auto_orient)
         @current_geometry.auto_orient
       end
-
+      
       @source_file_options = @source_file_options.split(/\s+/) if @source_file_options.respond_to?(:split)
       @convert_options     = @convert_options.split(/\s+/)     if @convert_options.respond_to?(:split)
 
--- a/lib/paperclip/thumbnail.rb
+++ b/lib/paperclip/thumbnail.rb
@@ -42,7 +42,6 @@ module Paperclip
       if @auto_orient && @current_geometry.respond_to?(:auto_orient)
         @current_geometry.auto_orient
       end
-      
       @source_file_options = @source_file_options.split(/\s+/) if @source_file_options.respond_to?(:split)
       @convert_options     = @convert_options.split(/\s+/)     if @convert_options.respond_to?(:split)
 
@@ -78,10 +77,9 @@ module Paperclip
         parameters = parameters.flatten.compact.join(" ").strip.squeeze(" ")
 
         success = convert(parameters, 
-          :source => "#{File.expand_path(src.path)}#{'[' + @frame_index.to_s + ']' unless animated?}", 
-          :dest => File.expand_path(dst.path)
-          )
-          
+                          :source => "#{File.expand_path(src.path)}#{'[' + @frame_index.to_s + ']' unless animated?}", 
+                          :dest => File.expand_path(dst.path),
+                          )
       rescue Cocaine::ExitStatusError => e
         raise Paperclip::Error, "There was an error processing the thumbnail for #{@basename}" if @whiny
       rescue Cocaine::CommandNotFoundError => e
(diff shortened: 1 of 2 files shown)


tute:
I like this. Can you please address Hound comments, while we are on it? Also, we can achieve the initialization of the default value with `options.fetch(:frame_index, 0)`.

tute:
Thanks for your work and contribution!

lib/paperclip/thumbnail.rb
```diff
@@ -118,4 +119,4 @@ def identified_as_animated?
       raise Paperclip::Errors::CommandNotFoundError.new("Could not run the `identify` command. Please install ImageMagick.")
     end
   end
-end
+end
```
jacobbullock: I'll fix this in an update, i removed the white space in the wrong part of the file

lib/paperclip/thumbnail.rb
```diff
@@ -76,7 +77,7 @@ def make
 
         parameters = parameters.flatten.compact.join(" ").strip.squeeze(" ")
 
-        success = convert(parameters, :source => "#{File.expand_path(src.path)}#{'[0]' unless animated?}", :dest => File.expand_path(dst.path))
+        success = convert(parameters, :source => "#{File.expand_path(src.path)}#{'[' + @frame_index.to_s + ']' unless animated?}", :dest => File.expand_path(dst.path))
```
jacobbullock: do you want me to put this on two lines?  didn't seem like the code was adhering to a max line count.  Also, Not sure the hash syntax here with having to wrap the frame_index in brackets. and only add this whole block based not he unless condition.  thoughts?

tute: While it's true that existing code doesn't consistently adhere to the style guides we want to follow, but new code should whenever possible. Suggested rewrite:

```ruby
success = convert(
  parameters,
  source: "#{File.expand_path(src.path)}#{'[' + @frame_index.to_s + ']' unless animated?}",
  dest: File.expand_path(dst.path),
)
```

(I would even extract a nicer private method for `source:`'s value, but that's not needed now).

tute:
Looking good! Can you please add a regression test for this? Also, can you address the new Hound comments? Thank you very much for sharing your work! 😃

lib/paperclip/thumbnail.rb
```diff
@@ -76,7 +77,11 @@ def make
 
         parameters = parameters.flatten.compact.join(" ").strip.squeeze(" ")
 
-        success = convert(parameters, :source => "#{File.expand_path(src.path)}#{'[0]' unless animated?}", :dest => File.expand_path(dst.path))
+        success = convert(parameters, 
+          :source => "#{File.expand_path(src.path)}#{'[' + @frame_index.to_s + ']' unless animated?}",
```
tute: Should the `frame_index` be there if it _is_ animated, rather than if it's not?

jacobbullock:
i'll try to wrap all of this up this week.  been busy launching this new product

tute:
@jacobbullock congrats on your launch! And don't forget your good PR! :)

lib/paperclip/thumbnail.rb
jacobbullock: @tute so this is interesting. I updated this line and it already had the 'unless animated?' check.  It used to specify a [0] index.  However in the tests I did, i do not pass in the animated param, and the class sets that value to true by default.  It does make sense that we should be only doing this if animated, but maybe this is a different bug?  any thoughts?

lib/paperclip/thumbnail.rb
```diff
@@ -76,7 +76,10 @@ def make
 
         parameters = parameters.flatten.compact.join(" ").strip.squeeze(" ")
 
-        success = convert(parameters, :source => "#{File.expand_path(src.path)}#{'[0]' unless animated?}", :dest => File.expand_path(dst.path))
+        success = convert(parameters, 
+                          :source => "#{File.expand_path(src.path)}#{'[' + @frame_index.to_s + ']' unless animated?}", 
+                          :dest => File.expand_path(dst.path),
+                          )
```
jacobbullock: @tute do you what they want here for alignment?

lib/paperclip/thumbnail.rb
jacobbullock: @tute one more thing, i looks like ANIMATED_FORMATS only includes gif right now, which may be why the animated? method was returning false for videos.  I can update that as well if we want to make this method work only when animated is true, which makes the most sense.

jacobbullock:
@tute should I add the test in spec/thumbnail_spec.rb?

lib/paperclip/thumbnail.rb
tute: > @tute one more thing, i looks like ANIMATED_FORMATS only includes gif right now, which may be why the animated? method was returning false for videos. I can update that as well if we want to make this method work only when animated is true, which makes the most sense.

This sound good to me

tute:
New tests in `spec/thumbnail_spec.rb` sound good. Also, can you please address Hound comments? Thank you!

tute:
ping @jacobbullock. Thank you!

spec/paperclip/thumbnail_spec.rb
```diff
@@ -480,6 +480,39 @@ def to_s
...
+      end
+    end
+    
+    context "with a specified frame_index out of bounds" do
+      before do
+        @thumb = Paperclip::Thumbnail.new(@file,
+                                          geometry: "50x50",
+                                          frame_index: 20,
+                                          format: :jpg,
+                                          )
```
jacobbullock: @tute can you help me out with what Hound wants here?  I have tried a few different options for aligning the ( ), but it keeps reporting errors

jacobbullock:
@tute I think the rest of these issues are just minor things.  Hound is reporting some things I don't understand, like it's asking me to use the hash syntax, when I am using the hash syntax.  

let me know if you want me to address anything else here.

bdewater:
Hound is asking you to use 1.9 hash syntax, e.g. `{ key: "value" }` instead of `{ key => "value" }`. The project itself is setting a bad example by still using 'hash rocket' syntax.

jacobbullock:
@bdewater tired eyes, I thought it was talking about the #{ } syntax in strings.  Thanks for clarifying. =)

tute:
Merged into master. Thank you! :)

PR #2169, opened 2016-04-07
Fixes cocaine path duplication bug

christophed:
If your path gets too big (e.g., you start appending your path thousands of times) you're going to have a bad time. On our boxes we noticed that paperclip would keep appending the Cocaine command line path, which crushes the CPU after some threshold.

Modified test to check for this and added a fix.


tute:
Thank you! I squashed and merged into master, keeping your authorship. 👍

PR #2173, opened 2016-04-15
Fixed case when file is not present on cloud

vimutter:


lib/paperclip/storage/fog.rb
```diff
@@ -170,6 +170,7 @@ def copy_to_local_file(style, local_dest_path)
         log("copying #{path(style)} to local file #{local_dest_path}")
         ::File.open(local_dest_path, 'wb') do |local_file|
           file = directory.files.get(path(style))
+          return false unless file
           local_file.write(file.body)
```
tute: I'd write this as:
```ruby
if file = directory.files.get(path(style))
  local_file.write file.body
end
```
Do you think it reads better?

tute:
This looks good to me. What was the behaviour before your fix? What is it now? Can you add that information to your commit message? Then we'll merge. Thanks!

lib/paperclip/storage/fog.rb
vimutter: @tute Usually we avoid assignment in conditionals as it always triggers internal question whether this line is typo or not. I.e. it requires check of surrounding of the line, which is sometimes ok. 
I.e. I can change this line if you wish so.

vimutter:
@tute Currently it fails with NoMethodError on nil when file is not there. And since `nil` is proper response from fog on missing files and this method already returns false if any error happen (see line 178) I decided that returning false here would be properly handled on client code.

lib/paperclip/storage/fog.rb
tute: > @tute Usually we avoid assignment in conditionals as it always triggers internal question whether this line is typo or not. I.e. it requires check of surrounding of the line, which is sometimes ok. 
> I.e. I can change this line if you wish so.

I don't feel very strongly either way, so following the Hound (who didn't bark at this line) is good with me. 👍

tute:
> Currently it fails with NoMethodError on nil when file is not there. And since nil is proper response from fog on missing files and this method already returns false if any error happen (see line 178) I decided that returning false here would be properly handled on client code.

Great! Can you please squash your commit together, rebase against latest master, and add this context to the commit message? Thanks!

tute:
> Can you please squash your commit together, rebase against latest master, and add this context to the commit message? Thanks!

Did it myself, and pushed to master. Thank you! 🎉

Issue #1432, opened 2014-02-07
Readme missing info on how to configure for testing


steakchaser:
I've seen a lot of different approaches being recommended for how one should work with Paperclip during testing, but there's nothing official being recommended in the Readme.  This would be very useful.

For example, prior to 4.0, I could essentially disable Paperclip from processing any attachments by doing this in my factory:

```ruby
factory :attachment do
  supporting_documentation_file_name { 'test.pdf' }
  supporting_documentation_content_type { 'application/pdf' }
  supporting_documentation_file_size { 1024 }
  # ...
end
```

This no longer works, and instead I get a validation error:
> ActiveRecord::RecordInvalid: Validation failed: Image translation missing: en.activerecord.errors.models.attachment.attributes.supporting_documentation.spoofed_media_type

There are quite a few options being recommended for 3.x (i.e. http://stackoverflow.com/questions/3294824/how-do-i-use-factory-girl-to-generate-a-paperclip-attachment).  There's almost nothing on this topic yet for 4.0.

If I come to a reasonable approach, I'll write it up, but would like feedback from others too.

maclover7:
Hi everybody! Is this still an issue for you in Paperclip; I know this issue is from approximately 1 year ago. If it is still an issue, can you please provide the code that's causing you the error? Thanks!

christophe-dufour:
As far as I'm concerned, I still can't see on the readme a recommended why to configure FactoryGirl with paperclip.

Should I use fixture_file_upload or not ??

blackjackcf:
@christophe-dufour - I've been using fixture_file_upload and then configuring my spec_helper to just clear out the contents of the folder that Paperclip uploads into. It's not pretty, but it works.

kulte:
@maclover7 like @ColinTheRobot says, it's impossible to figure out the configuration to get `#stub_paperclip_s3` and `#paperclip_fixture` to work. What context should these be called in? Have these worked for anyone ever?

kulte:
It would appear by the lack of dog-fooding here that perhaps these methods are dead code? Or am I way off base? https://github.com/thoughtbot/paperclip/blob/master/features/step_definitions/s3_steps.rb

dgilperez:
@blackjackcf could you please share some code about your approach? I keep banging my head against the wall trying to figure this out ...

blackjackcf:
@dgilperez - Here's my setup for my paperclip defaults:

    paperclip: &defaults
      storage: s3
[...]

PR #2183, opened 2016-04-30
Update README.md

asaganda:
Added FactoryGirl configuration details

### DIFF
--- a/README.md
+++ b/README.md
@@ -929,6 +929,17 @@ config.after(:suite) do
   FileUtils.rm_rf(Dir["#{Rails.root}/spec/test_files/"])
 end
 ```
+
+**Example of test configuration with Factory Girl**
+
+
+```ruby
+FactoryGirl.define do
+  factory :user do
+    avatar { File.new("#{Rails.root}/spec/support/fixtures/image.jpg") }
+  end
+end
+```
 ---
 
 Contributing


asaganda:
@vinvasir

README.md
```diff
@@ -929,6 +929,31 @@ config.after(:suite) do
...
+
+**Configurating Tests (Factory Girl)**
+
+FactoryGirl setup for the model that's getting the Paperclip attachment:
+
+```ruby
+FactoryGirl.define do
+  factory :asset do
+    attachment { Rack::Test::UploadedFile.new("#path_to_attachment", "attachment_file_type (e.g. image/png)") }
+  end
```
tute: Thanks! What do you think of a less generic example, like:

```ruby
factory :user do
  avatar { File.new("#{Rails.root}/spec/support/fixtures/image.jpg") }
end
```

(taken from https://stackoverflow.com/questions/3294824/how-do-i-use-factory-girl-to-generate-a-paperclip-attachment)

README.md
```diff
@@ -929,6 +929,31 @@ config.after(:suite) do
   FileUtils.rm_rf(Dir["#{Rails.root}/spec/test_files/"])
 end
 ```
+
+**Configurating Tests (Factory Girl)**
```
tute: I'd rephrase this to:

> **Configurating Tests with Factory Girl**

What do you think of that?

README.md
```diff
@@ -929,6 +929,31 @@ config.after(:suite) do
...
+
+FactoryGirl setup for the model that's getting the Paperclip attachment:
```
tute: If we change the title, maybe we don't need this sentence.

README.md
tute: Or:

> Example of test configuration with Factory Girl

README.md
```diff
@@ -929,6 +929,31 @@ config.after(:suite) do
...
+
+```ruby
+RSpec.configure do |config|
+  # ...
+  config.after(:suite) do
+     FileUtils.rm_rf(Dir["#path_where_attachments_are_saved"])
+  end
+  #...
+end
+```
```
tute: Images will be loaded in `tmp/`, which is always safe to remove, and can stay as a cache. I don't think we'll need this cleanup section.

tute:
Related issue: https://github.com/thoughtbot/paperclip/issues/1432

tute:
Thank you. 🚀

PR #2185, opened 2016-04-30
README elucidation for Ruby beginners

jmsoper:
Added links to referenced tools for beginners, blocked out User model further in Quick Start

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ In development mode, you might add this line to `config/environments/development
 Paperclip.options[:command_path] = "/usr/local/bin/"
 ```
 
-If you're on Mac OS X, you'll want to run the following with Homebrew:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -234,6 +235,10 @@ def create
   @user = User.create( user_params )
 end
 
+def show
+  @user = User.find( user_params )
+end
+
```
tute: Do we need to add this method here? If it doesn't add to the documentation, I'd drop it, leaving it as it was.

tute:
One small comment, looks good!

jmsoper:
I thought it might be clearer to have it in, since we implement a show view
later in quick start?

tute:
> I thought it might be clearer to have it in, since we implement a show view later in quick start?

That makes sense. What I'd do (and after the conversation with Vadim talking about how we infer the applicatino already exists) is to rename the `Show View` section to `View Helpers`.

PR #2194, opened 2016-05-05
Clarify semantics of the only_process option (documentation-only)

cfinucane:
This was really confusing to me, and I think this small change would help to clarify what's going on without having to trace through the code.


tute:
Great timesaver, thank you!

Issue #2199, opened 2016-05-10
tests on master should have no failures on any computer


tute:
Right now they are only green in CI. Different computers fail differently, in mine I get:

```
rspec ./spec/paperclip/validators_spec.rb:21 # Paperclip::Validators using the helper prevents you from attaching a file that violates that validation
rspec ./spec/paperclip/storage/s3_spec.rb:696 # Paperclip::Storage::S3 Parsing S3 credentials with a s3_host_name in them gets the right s3_host_name if the key does not exist
rspec ./spec/paperclip/attachment_processing_spec.rb:52 # Attachment Processing using validates_attachment processes attachments given a valid assignment
```

tute:
`rspec ./spec/paperclip/validators_spec.rb:21` seem to depend on the seed, the others two fail "more reliably".

pedrosmmoreira:
Hi @tute, happy to help out with this one if possible. Running the specs now

tute:
Thank you! 😃

pedrosmmoreira:
Thank you and all the maintainers for your time and effort :)

pedrosmmoreira:
`rspec ./spec/paperclip/storage/s3_spec.rb:696` is failing to create a new `Aws::Resource` on https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/storage/s3.rb#L247 as the test setup neither sets a `test` key nor an `s3_region` key. Since the Aws sdk does regex match on the passed in string, we fail on `NoMethodError`

I don't know how realistic this testing scenario: would not a failure to include the s3 hash in options passed to `has_attached_file` cause a fallback to whatever is defined in the environment? In that case, nothing being found, Paperclip would fallback to setting storage to `:filesystem`, correct?

Seems like we have a quick fix by just adding the test env hash with `s3_region` pointing to an empty string (in https://github.com/thoughtbot/paperclip/blob/master/spec/paperclip/storage/s3_spec.rb#L667). 

I can PR this if you agree that's a reasonable solution.

tute:
That is a very weird spec. Why do we need the `rails_env("test")` block?

> Seems like we have a quick fix by just adding the test env hash with s3_region pointing to an empty string (in https://github.com/thoughtbot/paperclip/blob/master/spec/paperclip/storage/s3_spec.rb#L667).
>
> I can PR this if you agree that's a reasonable solution.

That makes sense to me. 👍 Thanks for your help!

[...]

PR #2200, opened 2016-05-10
Correct failing S3 storage spec

pedrosmmoreira:

This spec was failing on obtaining a S3 instance with an `nil` region key. Since `Aws::Resource` does a regex match against the region key passed in, we would fail on `NoMethodError`.

This corrects the issue by setting a region key to an empty string in the test setup.


tute:
Thank you! ❤️

PR #2201, opened 2016-05-10
Correct attachment processing spec

pedrosmmoreira:

The attachment processing spec was failing since the setup required done in a before block, namely a call to `rebuild_class`, was running restricted to one of two contexts in the test file.
This was leaving the `Dummy` class in a bad state and affected other tests.

This PR moves the `before` block to run for the both of the contexts.


tute:
Thank you, again! 🎉 🎉 🎉

PR #2202, opened 2016-05-10
Correct validators spec

pedrosmmoreira:

The validators spec was failing since the setup required was missing a call to 'rebuild_class' and leaving the Dummy class in a bad state.

This PR adds the call to the before block setup.


tute:
Thanks so much for helping with this. 😃

PR #2210, opened 2016-05-11
The uri io adapter should use the content-disposition filename

netmask:
When a URI is passed to the model attachment  as the "attachment source"

example:

```ruby 
model.document = "http://examle.com/download_pdf"
```
paperclip will download this resource and set the last element of the path in this case, "download_pdf" as the file name, This PR allows the URI adapter to use the "content-disposition" header if is available and set that value as the original filename which in most cases are the expected one.

``` 
model.document = "http://examle.com/download_pdf"
# the response header contains the real name
# Content-disposition: attachment; filename=awesome_pdf_2016.pdf
# model.document.url => /storage/awesome_pdf_2016.pdf instead of download_pdf
```

### DIFF
--- a/lib/paperclip/io_adapters/.#data_uri_adapter.rb
+++ b/lib/paperclip/io_adapters/.#data_uri_adapter.rb
@@ -1,1 +1,0 @@
-netmask@Janet.35319
--- a/lib/paperclip/io_adapters/.#data_uri_adapter.rb
+++ b/lib/paperclip/io_adapters/.#data_uri_adapter.rb
@@ -1,0 +1,1 @@
+netmask@Janet.35319
--- a/lib/paperclip/io_adapters/uri_adapter.rb
+++ b/lib/paperclip/io_adapters/uri_adapter.rb
@@ -32,8 +32,8 @@ module Paperclip
     end
 
     def extract_attachment_filename
-      if @content.meta.has_key? 'content-disposition'
-        @original_filename = @content.meta['content-disposition']
+      if @content.meta.has_key? "content-disposition"
+        @original_filename = @content.meta["content-disposition"]
                                  .match(/filename="([^"]*)"/)[1]
       end
     end
--- a/lib/paperclip/io_adapters/uri_adapter.rb
+++ b/lib/paperclip/io_adapters/uri_adapter.rb
@@ -2,6 +2,8 @@ require 'open-uri'
 
 module Paperclip
   class UriAdapter < AbstractAdapter
+    attr_writer :content_type
+
     def initialize(target)
       @target = target
       @content = download_content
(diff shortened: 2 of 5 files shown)


lib/paperclip/io_adapters/uri_adapter.rb
```diff
@@ -28,6 +31,13 @@ def cache_current_values
       @size = @content.size
     end
 
+    def extract_attachment_filename
+      if @content.meta.has_key? "content-disposition"
+        @original_filename = @content.meta["content-disposition"]
+                                 .match(/filename="([^"]*)"/)[1]
+      end
+    end
```
tute: While we are here we can refactor some code. Suggestion:

```ruby
    def cache_current_values
      self.content_type = content_type_from_content || "text/html"

      self.original_filename =
        filename_from_content_disposition ||
        filename_from_path ||
        default_filename

      @size = @content.size
    end

    def content_type_from_content
      if @content.respond_to?(:content_type)
        @content.content_type
      end
    end

    def filename_from_content_disposition
      if @content.meta.has_key?("content-disposition")
        @original_filename = @content.meta["content-disposition"]
                                 .match(/filename="([^"]*)"/)[1]
      end
    end

    def filename_from_path
      @target.path.split("/").last
    end

    def default_filename
      "index.html"
    end
```

lib/paperclip/io_adapters/uri_adapter.rb
```diff
@@ -28,6 +31,13 @@ def cache_current_values
       @size = @content.size
     end
 
+    def extract_attachment_filename
+      if @content.meta.has_key? "content-disposition"
```
tute: Will `@content` always have the `meta` method defined?

tute:
Looks good! Thank you.

1. Left a comment on a potential refactor. What do you think of it?
1. Can you please add a note to the NEWS file?
1. If this is backwards compatible, do you think we should only apply it to the master branch, and not the v4.3 one?

Again, thanks for your contribution! 🎉

lib/paperclip/io_adapters/uri_adapter.rb
netmask: It should, the adapter uses the open uri module which always contains the meta hash

lib/paperclip/io_adapters/uri_adapter.rb
netmask: this looks much better 👍

netmask:
I think this should be applied also to v4.3 branch because is the latest stable version :)

lib/paperclip/io_adapters/uri_adapter.rb
```diff
@@ -2,30 +2,51 @@
...
-    def download_content
-      open(@target)
+      @size = @content.size
     end
 
-    def cache_current_values
-      @original_filename = @target.path.split("/").last
-      @original_filename ||= "index.html"
-      self.original_filename = @original_filename.strip
+    def content_type_from_content
```
tute: These four methods should be `private`.

tute:
> I think this should be applied also to v4.3 branch because is the latest stable version :)

But this changes behavior, and is thus backwards incompatible.

tute:
We can consider it a bugfix, though, indeed.

netmask:
Yes , we can considere it a bug, because the URI specifies the resource location, protocol but not the file name itself(?).

lib/paperclip/io_adapters/.#data_uri_adapter.rb
```diff
@@ -0,0 +1 @@
+netmask@Janet.35319
```
tute: It seems like this is a runaway file?

tute:
Sounds good to me. Left a single comment, this looks almost ready. Thank you! 🎉

lib/paperclip/io_adapters/.#data_uri_adapter.rb
netmask: ups yea, its an emacs backup file :P sorry about that

tute:
Merged into master. Thank you! :)

tute:
> Merged into master. Thank you! :)

I meant, merged into v4.3. Do you want to port this fix into master, too? Thank you!

netmask:
sure !!

PR #2225, opened 2016-06-10
User Paperclip namespace as Capistrano task

victorngkp:
Since the task `deploy` is quite exclusive to Capistrano, we should use Paperclip's own namespace instead, so not to be confused with Capistrano's `deploy`.


README.md
```diff
@@ -864,7 +864,7 @@ namespace :deploy do
   end
 end
 
-after("deploy:compile_assets", "deploy:build_missing_paperclip_styles")
+after("deploy:compile_assets", "paperclip:build_missing_paperclip_styles")
```
tute: Should this read `paperclip:build_missing_styles`, to avoid repetition?

PR #2227, opened 2016-06-11
Use Paperclip namespace as Capistrano task

victorngkp:
Since the task `deploy` is quite exclusive to Capistrano, we should use Paperclip's own namespace instead, so not to be confused with Capistrano's `deploy`.


victorngkp:
@tute Updated code based on your comment here - https://github.com/thoughtbot/paperclip/pull/2225#discussion_r66670638. I had to close that PR because I realized I accidentally opened a PR from my fork master. Thanks!

tute:
Thank you!

PR #2229, opened 2016-06-14
Make fingerprints configurable to prepare for a new default in the future

bdewater:
This is the continuation of the work by @medcat in #1708. I've rebased his work on master and made separate commits addressing changes requested in that pull request. If the maintainers of the gem give the green light I will squash the commits (preserving authorship) for clean commit history 😄

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,22 @@
 master:
 
+* Improvement: make the fingerprint digest configurable per attachment. The
+  default remains MD5 but this will change in a future version because it is
+  not considered secure anymore against intentional file corruption. For more
+  info, see https://en.wikipedia.org/wiki/MD5#Security
+
+  You can change the digest used for an attachment by adding the
+  `:adapter_options` parameter to the `has_attached_file` options like this:
+  `has_attached_file :avatar, adapter_options: { hash_digest: Digest::SHA256 }`
+
+  Use the rake task to regenerate fingerprints with the new digest for a given
+  class. Note that this does **not** check the file integrity using the old
+  fingerprint. Run the following command to regenerate fingerprints for all
+  User attachments:
+  `CLASS=User rake paperclip:refresh:fingerprints`
+  You can optionally limit the attachment that will be processed, e.g:
+  `CLASS=User ATTACHMENT=avatar rake paperclip:refresh:fingerprints`
+
 5.1.0 (2016-08-19):
 
 * Add default `content_type_detector` to `UploadedFileAdapter` (#2270)
--- a/README.md
+++ b/README.md
@@ -737,10 +737,10 @@ is specified in `:hash_data`. The default value for `:hash_data` is `":class/:at
 
 For more on this feature, read [the author's own explanation](https://github.com/thoughtbot/paperclip/pull/416)
 
-MD5 Checksum / Fingerprint
+Checksum / Fingerprint
 -------
 
-An MD5 checksum of the original file assigned will be placed in the model if it
+A checksum of the original file assigned will be placed in the model if it
 has an attribute named fingerprint.  Following the user model migration example
 above, the migration would look like the following:
 
@@ -755,6 +755,17 @@ class AddAvatarFingerprintColumnToUser < ActiveRecord::Migration
   end
 end
 ```
+
+The algorithm can be specified using a configuration option; it defaults to MD5
+for backwards compatibility with Paperclip 5 and earlier.
+
+```ruby
+has_attached_file :some_attachment, adapter_options: { hash_digest: Digest::SHA256 }
+```
+
+Run `CLASS=User ATTACHMENT=avatar rake paperclip:refresh:fingerprints` after
+changing the digest on existing attachments to update the fingerprints in the
+database.
 
 File Preservation for Soft-Delete
 -------
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -33,6 +33,7 @@ module Paperclip
         :use_timestamp         => true,
         :whiny                 => Paperclip.options[:whiny] || Paperclip.options[:whiny_thumbnails],
         :validate_media_type   => true,
+        :adapter_options       => { hash_digest: Digest::MD5 },
         :check_validity_before_processing => true
       }
     end
@@ -97,7 +98,8 @@ module Paperclip
     # attachment:
     #   new_user.avatar = old_user.avatar
     def assign(uploaded_file)
-      @file = Paperclip.io_adapters.for(uploaded_file)
+      @file = Paperclip.io_adapters.for(uploaded_file,
+                                        @options[:adapter_options])
       ensure_required_accessors!
       ensure_required_validations!
 
@@ -523,15 +525,18 @@ module Paperclip
       begin
         raise RuntimeError.new("Style #{name} has no processors defined.") if style.processors.blank?
         intermediate_files = []
+        original = @queued_for_write[:original]
 
-        @queued_for_write[name] = style.processors.inject(@queued_for_write[:original]) do |file, processor|
+        @queued_for_write[name] = style.processors.
+          reduce(original) do |file, processor|
           file = Paperclip.processor(processor).make(file, style.processor_options, self)
           intermediate_files << file unless file == @queued_for_write[:original]
           file
         end
 
         unadapted_file = @queued_for_write[name]
-        @queued_for_write[name] = Paperclip.io_adapters.for(@queued_for_write[name])
+        @queued_for_write[name] = Paperclip.io_adapters.
+          for(@queued_for_write[name], @options[:adapter_options])
         unadapted_file.close if unadapted_file.respond_to?(:close)
         @queued_for_write[name]
       rescue Paperclip::Errors::NotIdentifiedByImageMagickError => e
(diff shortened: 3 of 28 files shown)


bdewater:
@tute I've again rebased on master now that 5.0 is released. I assume it is too late now to change the default - following semantic versioning that would be 6.0. Could we work on getting the changes in with MD5 as default so that changing the default in the future is only a small change?

tute:
> Could we work on getting the changes in with MD5 as default so that changing the default in the future is only a small change?

This makes perfect sense to me. Thank you!

bdewater:
Rebased again @tute !

NEWS
```diff
@@ -1,6 +1,22 @@
 master:
 * (port from 4.3) Improvement: the `URI adapter` now uses the content-disposition header to name the downloaded file.
 
+* Improvement: make the fingerprint digest configurable per attachment. The
+  default remains MD5 but this will change in a future version because it is
+  not considered secure anymore against intentional file corruption.
```
tute: Can you gather resources on this assertion, so we provide more information to paperclip users?

README.md
```diff
@@ -745,6 +745,16 @@ class AddAvatarFingerprintColumnToUser < ActiveRecord::Migration
 end
 ```
 
+The algorithm can be specified using a configuration option; it defaults to MD5.
```
tute: I'd add here "for backwards compatibility", so people know we are working on providing better defaults already.

lib/paperclip/io_adapters/abstract_adapter.rb
```diff
@@ -8,8 +8,20 @@ class AbstractAdapter
...
 
+    def initialize(target, options = {})
+      @target = target
+      @options = options
+    end
+
     def fingerprint
-      @fingerprint ||= Digest::MD5.file(path).to_s
+      @fingerprint ||= begin
+        digest = @options.fetch(:hash_digest, Digest::MD5).new
```
tute: Do we need to specify the default, if we already set it when paperclip starts?

lib/paperclip/io_adapters/identity_adapter.rb
```diff
@@ -1,8 +1,11 @@
 module Paperclip
   class IdentityAdapter < AbstractAdapter
-    def new(adapter)
+    def new(adapter, _)
       adapter
     end
+
+    def initialize
```
tute: What does this definition with no arguments do?

lib/paperclip/io_adapters/uri_adapter.rb
```diff
@@ -4,7 +4,8 @@ module Paperclip
   class UriAdapter < AbstractAdapter
     attr_writer :content_type
 
-    def initialize(target)
+    def initialize(target, options = {})
+      super
       @target = target
```
tute: `super` sets `@target`, we can drop this line.

lib/paperclip/tempfile_factory.rb
```diff
@@ -13,7 +13,7 @@ def extension
     end
 
     def basename
-      Digest::MD5.hexdigest(File.basename(@name, extension))
+      Digest::SHA1.hexdigest(File.basename(@name, extension))
```
tute: Should this be the configuration option, or be rollbacked to `MD5` for backwards compatibility?

lib/tasks/paperclip.rake
```diff
@@ -64,7 +64,8 @@ namespace :paperclip do
       names = Paperclip::Task.obtain_attachments(klass)
       names.each do |name|
         Paperclip.each_instance_with_attachment(klass, name) do |instance|
-          if file = Paperclip.io_adapters.for(instance.send(name))
+          attachment = instance.send(name)
+          if file = Paperclip.io_adapters.for(instance.send(name), attachment.options[:adapter_options])
```
tute: We can use `attachment` instead of `instance.send(name)` as first argument.

lib/tasks/paperclip.rake
```diff
@@ -109,7 +123,20 @@ namespace :paperclip do
     end
   end
 
- desc "find missing attachments. Useful to know which attachments are broken"
+  desc "Upgrades the MD5 fingerprints to the SHA1 fingerprints."
```
tute: Do you mean `SHA256` instead of `SHA1`? In any case, where do you set the option?

spec/paperclip/io_adapters/attachment_adapter_spec.rb
```diff
@@ -41,8 +42,8 @@
       assert ! @subject.nil?
     end
 
-    it "generates a MD5 hash of the contents" do
-      expected = Digest::MD5.file(@file.path).to_s
+    it "generates a SHA1 hash of the contents" do
+      expected = Digest::SHA1.file(@file.path).to_s
```
tute: `SHA256`?

tute:
This PR looks great, thank you for your great work, @bdewater! Left a few comments. Can you please address the ones you agree with on different commits, so as to ease the code review?

I might not be getting something around the `SHA1` vs `SHA256`. In that case, please explain why we have both.

Thank you!

lib/paperclip/io_adapters/identity_adapter.rb
bdewater: I'll have to look at this closer later because at first glance I don't get what this class does. Changing it consistently breaks "Paperclip::Storage::S3 An attachment that uses S3 for storage and has a question mark in file name returns a replaced version for {path|url}" specs.

lib/paperclip/io_adapters/abstract_adapter.rb
bdewater: I've removed the duplication from Paperclip::Attachment.

lib/paperclip/attachment.rb
```diff
@@ -33,7 +33,7 @@ def self.default_options
         :use_timestamp         => true,
         :whiny                 => Paperclip.options[:whiny] || Paperclip.options[:whiny_thumbnails],
         :validate_media_type   => true,
-        :adapter_options       => { hash_digest: Digest::MD5 },
+        :adapter_options       => { },
```
tute: This hash contains all the default options. For documentation purposes, I'd leave the default here, and not in code as it was.

lib/paperclip/io_adapters/abstract_adapter.rb
tute: > I've removed the duplication from Paperclip::Attachment.

Let's remove this instance, and leave the one in the defaults hash for better code documentation. Thank you! 😃

spec/paperclip/io_adapters/data_uri_adapter_spec.rb
```diff
@@ -20,7 +20,7 @@
   context "a new instance" do
     before do
       @contents = "data:image/png;base64,#{original_base64_content}"
-      @subject = Paperclip.io_adapters.for(@contents)
+      @subject = Paperclip.io_adapters.for(@contents, hash_digest: Digest::SHA1)
```
tute: Should we use MD5 or SHA256 here?

tute:
Thanks for your fast and great work, @bdewater! Left a comment around where to put the default hash function. Also, why do we mention SHA1 in the tests? Should we use the previous, or the new one?

bdewater:
When moving the default hash location those S3 specs break in the same way as with the identity adapter 😠 I'll try some more in the weekend.

bdewater:
I believe I've addressed all your comments @tute, I'll squash and rebase again if you agree it's ready for merging.

tute:
Left one question. Please do squash and rebase. Thank you so much!

lib/paperclip/io_adapters/identity_adapter.rb
```diff
@@ -1,12 +1,12 @@
 module Paperclip
   class IdentityAdapter < AbstractAdapter
-    def new(adapter)
+    def self.new(adapter, _)
```
tute: Why do we call this `adapter`, but `target` in other places?

tute:
Looking good. After we fix `IdentityAdapter` and make it look like the other adapters we can merge. Thanks a ton. :)

bdewater:
Done :)!

lib/paperclip/io_adapters/identity_adapter.rb
```diff
@@ -1,12 +1,12 @@
 module Paperclip
   class IdentityAdapter < AbstractAdapter
-    def new(adapter)
-      adapter
```
tute: This is very confusing to me. Do we need this identity adapter? Do we need it to have a `new` class or instance method? The current PR changes how the Ruby language works (`.new` initializes and returns an instance), and that is not right. What happens if we drop this adapter?

bdewater: Perhaps @jyurek could shine some light on what this adapter is supposed to do because it confused me as well.

bdewater: I've reverted overriding `#new` in favor of the original implementation (overriding initialize to take no arguments and new as an instance method), I agree that it is less confusing. I've poked around a bit but it seems that one would need to refactor a lot to get rid of this adapter. That seems out of scope for this PR.

tute:
This is a great change. Thanks so much for all your work! 🎉

PR #2230, opened 2016-06-14
replaced 3 if-not-blank checks with if-present checks

humzashah:
`value.present?` is more readable than `!value.blank?`. following this, 3 instances of `!value.blank?` have been replaced with `value.present?`


tute:
Thank you!

Issue #1923, opened 2015-07-05
Net::ReadTimeout error


Brinatti:
I'm trying to upload a file from URL but getting error Net::ReadTimeout.

I'm doing like described in the following link:
https://github.com/thoughtbot/paperclip/wiki/Attachment-downloaded-from-a-URL

jyurek:
Can you access those links directly, not through paperclip?

tute:
Closing due to lack of activity. Happy to reopen if we find steps to reproduce this problem.

droptheplot:
@tute @jyurek It happens if file from URL couldn't be reached within default timeout. I think if there was a way to change [read_timeout](http://ruby-doc.org/stdlib-2.1.1/libdoc/net/http/rdoc/Net/HTTP.html#attribute-i-read_timeout) it'll help.

```shell
File "...lib/ruby/2.2.0/net/protocol.rb" line 158 in rescue in rbuf_fill
File "...lib/ruby/2.2.0/net/protocol.rb" line 152 in rbuf_fill
File "...lib/ruby/2.2.0/net/protocol.rb" line 134 in readuntil
File "...lib/ruby/2.2.0/net/protocol.rb" line 144 in readline
File "...lib/ruby/2.2.0/net/http/response.rb" line 39 in read_status_line
File "...lib/ruby/2.2.0/net/http/response.rb" line 28 in read_new
File "...lib/ruby/2.2.0/net/http.rb" line 1414 in block in transport_request
File "...lib/ruby/2.2.0/net/http.rb" line 1411 in catch
File "...lib/ruby/2.2.0/net/http.rb" line 1411 in transport_request
File "...lib/ruby/2.2.0/net/http.rb" line 1384 in request
File "...lib/ruby/2.2.0/open-uri.rb" line 324 in block in open_http
File "...lib/ruby/2.2.0/net/http.rb" line 853 in start
File "...lib/ruby/2.2.0/open-uri.rb" line 318 in open_http
File "...lib/ruby/2.2.0/open-uri.rb" line 736 in buffer_open
File "...lib/ruby/2.2.0/open-uri.rb" line 211 in block in open_loop
File "...lib/ruby/2.2.0/open-uri.rb" line 209 in catch
File "...lib/ruby/2.2.0/open-uri.rb" line 209 in open_loop
File "...lib/ruby/2.2.0/open-uri.rb" line 150 in open_uri
File "...lib/ruby/2.2.0/open-uri.rb" line 716 in open
[...]

PR #2232, opened 2016-06-15
Add read_timeout option to UriAdapter download_content method

droptheplot:
`Net::ReadTimeout` exception will occur if remote file couldn't be reached within default timeout.
This PR adds option `:read_timeout` to Paperclip to override default `open` timeout from UriAdapter download_content method.

#1923

### DIFF
--- a/lib/paperclip.rb
+++ b/lib/paperclip.rb
@@ -98,7 +98,7 @@ module Paperclip
       :swallow_stderr => true,
       :content_type_mappings => {},
       :use_exif_orientation => true,
-      :read_timeout => 120
+      :read_timeout => nil
     }
   end
 
--- a/lib/paperclip.rb
+++ b/lib/paperclip.rb
@@ -97,7 +97,8 @@ module Paperclip
       :log_command => true,
       :swallow_stderr => true,
       :content_type_mappings => {},
-      :use_exif_orientation => true
+      :use_exif_orientation => true,
+      :read_timeout => 120
     }
   end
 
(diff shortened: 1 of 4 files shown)


lib/paperclip.rb
```diff
@@ -97,7 +97,8 @@ def self.options
       :log_command => true,
       :swallow_stderr => true,
       :content_type_mappings => {},
-      :use_exif_orientation => true
+      :use_exif_orientation => true,
+      :read_timeout => 120
```
tute: Where do we get the `120` default value from?

droptheplot: > Where do we get the 120 default value from?

You're right, I changed it to nil.

> Put a comma after the last item of a multiline hash.

Actually, trailing commas are not preferred: https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas.

tute: > Actually, trailing commas are not preferred: https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas.

It's fine, as we are not making it worse in this PR, but paperclip subscribes to thoughtbot guidelines, encoded in Hound CI: https://github.com/thoughtbot/guides/blob/master/style/ruby/sample.rb#L46-L59.

tute:
Great! Want to add a note in the NEWS file documenting your addition? Then we'll merge. Thanks!

droptheplot:
Done.

tute:
Thanks! 👍

PR #2234, opened 2016-06-17
on after_save and before_destroy hooks use public methods from attachment object

ssinghi:
The reason why I want this change is because we are writing a paperclip plugin, and we want to override the on attachment deletion behavior. Adding a new behavior to `queue_all_for_delete` just does not seems correct in our plugin.

I think there is no functional difference between calling `clear` or `queue_all_for_delete` for `before_destroy` AR hook, rather it should make the paperclip code simpler.


sidraval:

PR #2243, opened 2016-07-07
Update accepted server_side_encryption value

nickrivadeneira:
AWS no longer accepts `:aes256` as an encryption value and will raise
`Aws::S3::Errors::InvalidArgument` with the message `The encryption method
specified is not supported`. It should instead be `"AES256"`.

This commit makes no implementation change, but for those that use
source code and tests as documentation, it should help prevent
confusion.


tabfugnic:
LGTM

tute:
Thank you!

PR #2245, opened 2016-07-08
Assign file attributes before only_process

passabilities:
There could be a better way to do this, but this seems to be working.

I am using the `only_process` attribute in conjunction with [Delayed Paperclip](https://github.com/jrgifford/delayed_paperclip/) to process some files in the background. When I started using DP, I was getting errors saying that the attributes associated with the uploaded file (e.i. _content_type) were not set.

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -102,6 +102,7 @@ module Paperclip
       ensure_required_validations!
 
       if @file.assignment?
+        assign_attributes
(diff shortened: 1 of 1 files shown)


tute:
This change is backwards incompatible. Could you solve your issue some other way? Are there alternative solutions to your issue? It seems people use delayed_paperclip without the need for this change, what is different in your app?

Closing, as we won't merge, but happy to continue further discussion.

Thank you!

PR #2250, opened 2016-07-14
[backport] The uri io adapter should use the content-disposition filename

netmask:
@tute port from https://github.com/thoughtbot/paperclip/pull/2210

- The uri io adapter now seeks for the content-disposition header
  if this is pressent the value filename is taken instead of the
  last path segment for the resource file name
- Fix style comments
- Applied the Tute Costa refactor to URI Adapter.
- Added entry to  the NEWS file.
- Removed editor tracking file
- Fix test cases


tute:
Thank you very much! 👏

PR #2256, opened 2016-07-18
Rely on URI content_type only when provided. Fix #2118

### LINKED ISSUE #2118: uri_adapter.rb and content type
Hi there,

when you assign a URI as the source of a paperclip attachment, paperclip seems to rely on the remote server's content type for determining the attachments content type instead of the actual file's content. Is this intended? We stumbled upon this when a remote server we downloaded images from reports "application/octet-stream" as the content type. 

https://github.com/thoughtbot/paperclip/blob/master/lib/paperclip/io_adapters/uri_adapter.rb#L25

```
@content_type = @content.content_type if @content.respond_to?(:content_type)
@content_type ||= "text/html"
```

`@content` is the return value of a call to open(URI(...)) which returns a Tempfile. The Tempfile instance returns "application/octet-stream" when you call `#content_type`.

If this is something you think should be fixed, I can dig into it and provide a PR.

j15e:
As described in #2118, OpenURI::Meta#content_type will return `application/octet-stream` when no `content-type` header is set, so the intended `text/html` fallback isn't working.

Using the original meta fix this issue.

PS. I think we should rely on the ContentTypeDetector here unless we are sure the source(s) of your URIs are always trustworthy? Since the file is downloaded for processing anyway, I think it would be better to only rely on the ContentTypeDetector. 

But I am not aware of every use case, maybe it would be more trouble. Thought? Using the ContentTypeDetector here would fix my original issue (this PR only fix the current implementation).


j15e:
A bit more details... I think the original spirit of this adapter might not be well document and therefore not always used in the expected way.

I used it to download files that are directly uploaded to Amazon S3 by a frontend client, but I am not sure this would fit the original case scenario. With content uploaded by user on a S3 bucket, the content-type is not always set properly if there is no extension, therefore relying only on S3 to provide the content type is not very reliable. Also S3 don't check the file content as ContentTypeDetector can do.

I think I will simply created my own / an other adapter that will use the ContentTypeDetector no matter what the response headers say. Maybe it would be a saner behavior for the UrlAdapter, but what do you think?

edalford11:
Any updates on this? This is definitely still an issue with s3 in the latest version. There are many options suggest for allowing paperclip to make an exception for `application/octet-stream` but that does seem like its compromising the validation itself.

mike-burns:
Merged as f384174392ce192c9d76bd447902fe8a3ecf70ad, thank you!

Issue #2260, opened 2016-07-21
[Feature Request] support for background processing


morgoth:
Since Paperclip now requires Rails >= 4.2 it means that it has access to ActiveJob.

Will you consider integration with AJ for processing files in the background?
Currently it's possible using https://github.com/jrgifford/delayed_paperclip but it has lot of hacks, so it's not easy to maintain it.

Not sure how big effort would it be to get it into paperclip, but the first question is - is this something that you would like to have or not?

andreiglingeanu:
@morgoth +1!

tute:
@morgoth this is a great idea. What would it take to make paperclip ActiveJob aware?

morgoth:
@tute I created simple Proof of Concept in https://github.com/thoughtbot/paperclip/pull/2261 - we can move discussion there. I would like to get your feedback.

tute:
Closing in favor of PR. Thanks!

PR #2261, opened 2016-07-23
Background processing

morgoth:
Very basic implementation for background processing support using ActvieJob.

It requires more work, but I would like to start collecting your feedback about implementation and possible features.

I tested it with file and s3 storages locally.

Questions:
- ~~I think it make sense to add db column where we can mark that file is being processed () - do you think boolean `:attachment_name:_processing_in_background` is fine?~~ done
- To force processing in background one have to set array of styles in `process_in_background` options, but also set `only_processing` to fake `[:none]`, as empty array processes all by default. I would like to solve it some better way. How it should behave?

[ref #2260]

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -53,7 +53,8 @@ module Paperclip
     # +styles+ - a hash of options for processing the attachment. See +has_attached_file+ for the details
     # +only_process+ - style args to be run through the post-processor. This defaults to the empty list (which is
     #                  a special case that indicates all styles should be processed)
-    # +process_in_background+ - an array of styles to be processed in background (requires "active_job")
+    # +process_in_background+ - an array of styles to be processed in background
+    #                           (requires "active_job")
     # +default_url+ - a URL for the missing image
     # +default_style+ - the style to use when an argument is not specified e.g. #url, #path
     # +storage+ - the storage mechanism. Defaults to :filesystem
@@ -392,6 +393,14 @@ module Paperclip
       flush_writes
     end
 
+    def clear_enqueue_background_processing
+      @enqueue_background_processing = false
+    end
+
+    def enqueue_background_processing?
+      @enqueue_background_processing
+    end
+
     private
 
     def path_option
@@ -439,6 +448,7 @@ module Paperclip
       assign_file_information
       assign_fingerprint { @file.fingerprint }
       assign_timestamps
+      assign_background_processing
     end
 
     def assign_file_information
@@ -459,6 +469,12 @@ module Paperclip
       end
 
       instance_write(:updated_at, Time.now)
+    end
+
+    def assign_background_processing
+      if @options[:process_in_background].any?
+        @enqueue_background_processing = true
+      end
     end
 
     def post_process_file
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -21,6 +21,7 @@ module Paperclip
         :hash_digest           => "SHA1",
         :interpolator          => Paperclip::Interpolations,
         :only_process          => [],
+        :process_in_background => [],
         :path                  => ":rails_root/public:url",
         :preserve_files        => false,
         :processors            => [:thumbnail],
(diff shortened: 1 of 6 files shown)


lib/paperclip/attachment.rb
```diff
@@ -383,6 +386,21 @@ def instance_read(attr)
       end
     end
 
+    def generate_style_files(*styles)
+      @file = Paperclip.io_adapters.for(self)
+      @queued_for_write[:original] = @file
```
morgoth: It seems to be little hacky, but without this other styles won't process

lib/paperclip/has_attached_file.rb
```diff
@@ -100,6 +101,20 @@ def add_active_record_callbacks
       end
     end
 
+    def add_background_processing
+      require "paperclip/process_job"
```
morgoth: I added require in this place, to not force `active_job` presence for people that won't use this feature

lib/paperclip/has_attached_file.rb
```diff
@@ -100,6 +101,20 @@ def add_active_record_callbacks
...
+      name = @name
+      if @klass.respond_to?(:after_commit)
```
morgoth: Should I care for case when there is no `after_commit` and add `after_save` then?

spec/paperclip/attachment_spec.rb
```diff
@@ -1477,6 +1478,53 @@ def call(filename)
...
+  context "an attachment with process_in_background set" do
+    before do
+      ActiveJob::Base.queue_adapter = :test
+      ActiveJob::Base.logger = nil
+
+      rebuild_model styles: { thumb: "100x100" },
+                    only_process: [:none],
+                    process_in_background: [:thumb]
+      @file = File.new(fixture_file("5k.png"), "rb")
+      GlobalID.app = "paperclip-test"
```
morgoth: I couldn't make it work without it. In normal Rails app, this is set up by railties

spec/paperclip/attachment_spec.rb
```diff
@@ -1477,6 +1478,53 @@ def call(filename)
     end
   end
 
+  context "an attachment with process_in_background set" do
+    before do
+      ActiveJob::Base.queue_adapter = :test
+      ActiveJob::Base.logger = nil
+
+      rebuild_model styles: { thumb: "100x100" },
+                    only_process: [:none],
```
morgoth: This is a hack to skip processing in real time. We need to think about better UI, so user can decide which styles he wants to be processed in real time/background

spec/paperclip/attachment_spec.rb
```diff
@@ -1477,6 +1478,53 @@ def call(filename)
...
+      job = jobs[0]
+      assert_equal Paperclip::ProcessJob, job[:job]
+      assert_equal @dummy, GlobalID::Locator.locate(job[:args][0].values[0])
+      assert_equal "avatar", job[:args][1]
+      assert_equal "default", job[:queue]
+    end
+
+    it "does not process given styles in real time" do
+      attachment = @dummy.avatar
+      attachment.expects(:post_process).with(:none)
```
morgoth: I'm little confused when to use rspec and when minitest - which one is preferred?

morgoth:
I'm also wondering how configuration for future features should look like, in example:
`has_attached_file :avatar, process_in_background: [:thumb], background_job_queue_name: "paperclip,  ...` or maybe group them all under one key?
`has_attached_file :avatar, process_in_background: {styles: [:thumb], queue_name: "paperclip"}}`

andreiglingeanu:
@morgoth Let me know if I can help with something. Little bit of testing, anything. I'll be in touch.

ScotterC:
As a maintainer of DelayedPaperclip, I'd love for this to be in Paperclip core. Much less hacky and easier to maintain in the future.

morgoth:
@tute Do you think we can move it forward?
I think there are 2 blockers:
https://github.com/thoughtbot/paperclip/pull/2261#discussion_r71974153
https://github.com/thoughtbot/paperclip/pull/2261#issuecomment-234735329

bdewater:
How would callbacks work if this is integrated? I have a potential use case that looks quite a lot like the one described here: https://github.com/jrgifford/delayed_paperclip/issues/176

spec/paperclip/attachment_spec.rb
bdewater: RSpec, see message of 1f3a74678f459dc1b2d60efd49f3cfe4addc6f63

Tonkpils:
any update on this PR?

sidraval:

Issue #1445, opened 2014-02-18
Paperclip not deleting attachments after model is destroyed in test environment


tomrossi7:
I have a Rails 4 app and when I upgrade from Paperclip 3.5.2 to any of the newer versions, my attachments are no longer deleted from the filesystem after the associated model is destroyed. I can't see anything in the documentation that seems to have changed?

Here is my model code:

```
  has_attached_file :artwork, :styles => {:large => ["1400x1400#", :jpg],
                                          :cover => ["600x600#", :jpg],
                                          :medium => "170x170#",
                                          :thumb =>  "50x50#"},
                              :convert_options => {:all => "-quality 80 -colorspace RGB"}
```

Anything look incorrect?

momochi000:
@nateq314 I think I ran into the same problem as you and after some googling, came across this thread.  I wasted about 2 hours on this so for anyone else who comes here, the problem was you have 
`:preserve_files => "false"`
with false as a string.  Looking at the source it needs the :preserve_files option to be a boolean.  

The Readme is misleading here because it says
`has_attached_file :some_attachment, {
    :preserve_files => "true",
}`

which led me to believe it expected a string argument.  The string would evaluate to true so it would work in this case.  I know it says it defaults to false but I wanted to be explicit in case I forgot later on (which is inevitable for me =P )

maclover7:
Hi everybody! Is this still an issue for you in Paperclip; I know this issue is from approximately 1 year ago. If it is still an issue, can you please provide the code that's causing you the error? Thanks!

pierallard:
Hi everybody. I always have this issue with Paperclip and Rspec too. Attachments were not deleted in test environment, but they are in production and development environments.
Paperclip 4.2.1 and rspec 2.99.

Regards

henners66:
Just for the record I have run into the same issue when testing with RSpec with DatabaseCleaner - files are not being deleted on destroy of a model with an attachment.

I tried the solution using metadata suggested by @russellsilva (updated for latest RSpec - requires using RSpec.current_example instead of example) and although the strategy was correctly set it didn't help at all.

I then tried the suggestion by @mjy and explicitly called the commit callback and that worked.

johnloringpollard:
@pierallard I'm seeing exactly as you described. Is there a setting you set to make it delete the files in your test environment too?

mike-burns:
[...]

PR #2262, opened 2016-07-23
Update README preserve_files to be a boolean

unixmonkey:
Before this change, the value for `preserve_files` was the string `"true"`, potentially confusing users that they should use stringified booleans like `"true"` and `"false"`.

However `"false"` would evaluate as truthy, and preserve files.
This appears to have been a factor in https://github.com/thoughtbot/paperclip/issues/1445#issuecomment-51430186


tute:
This was weird. Thank you!

PR #2266, opened 2016-07-27
Don't allow trailing newlines in various checks.

benpickles:
There's a subtle difference between what `\Z` and `\z` consider the "end of string" which is that the uppercase version allows a single trailing newline:

```ruby
/\Afoo\Z/.match("foo\n")
# => #<MatchData "foo">

/\Afoo\Z/.match("foo\n\n")
# => nil

/\Afoo\z/.match("foo\n")
# => nil
```

The current usage of `\Z` / `\z` in the project can be discovered with:

```
$ git grep -iI "\\\z"
```

I've left the Cucumber steps alone as changing them makes adding the `has_attached_file` line fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ Quick Start
 ```ruby
 class User < ActiveRecord::Base
   has_attached_file :avatar, styles: { medium: "300x300>", thumb: "100x100>" }, default_url: "/images/:style/missing.png"
-  validates_attachment_content_type :avatar, content_type: /\Aimage\/.*\Z/
+  validates_attachment_content_type :avatar, content_type: /\Aimage\/.*\z/
 end
 ```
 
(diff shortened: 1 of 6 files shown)


tute:
Thanks! Why do we not want to allow for the new line character? Can there be a situation in which the headers do come with that character, and we don't process them properly after this change?

tute:
ping @benpickles. Thanks!

benpickles:
Sorry! This totally slipped my mind. 

Yes you're right that's where the risk lies in this pull request - though no tests fail and I haven't encountered any issues. I haven't exhaustively confirmed all the code paths that spit out mine types (in fact I haven't investigated any :p) but I think I'm right in saying that the change that could cause unexpected consequences is in the validation suggested to users in the README and not in the library itself - no wait, [here's one](https://github.com/thoughtbot/paperclip/pull/2266/files#diff-4fa49d3874cbdd5654c0d5518f54436fR144).

Either way the rest of the changes are probably what was originally intended in that a subdomain or file path cannot include newlines - so this could prevent potentially difficult-to-debug user error (though I imagine nobody has ever encountered this!).

tute:
Thanks!

PR #2270, opened 2016-08-01
Default to Paperclip::ContentTypeDetector

ivanoblomov:
This patch defaults `Paperclip::UploadedFileAdapter.content_type_detector` to `Paperclip::ContentTypeDetector`.

Without it, Paperclip allows the upload of any content, _including executables_, so long as the client forges a permitted content-type.

Security holes shouldn't be the default setting where we can avoid it.


tute:
Thank you very much, @ivanoblomov.

Issue #2271, opened 2016-08-03
Reduced redundancy AWS S3 storage class with Paperclip 5.0.0


codyrobbins:
[These lines](https://github.com/thoughtbot/paperclip/blob/95acf3b898bd782f4429d42fd2d8812505390328/lib/paperclip/storage/s3.rb#L101-L112) in the documentation indicated that the value for the `s3_storage_class` configuration option should be `:reduced_redundancy`. However, using this value with Paperclip 5.0.0 results in a `Aws::S3::Errors::InvalidStorageClass` error with the message `The storage class you specified is not valid`.

The [documentation for the `Aws::S3::Object#storage_class` method](http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html#storage_class-instance_method) in the `aws-sdk` gem documentation indicates that the correct value is now `REDUCED_REDUNDANCY` (along with a few other potential values).

The documentation should either be updated to reflect this fact or Paperclip itself updated to transform the value of the `s3_storage_class` configuration option to uppercase. I’m happy to make this change in a pull request but not sure which route you prefer.

tute:
> The documentation should either be updated to reflect this fact or Paperclip itself updated to transform the value of the s3_storage_class configuration option to uppercase. I’m happy to make this change in a pull request but not sure which route you prefer.

I'd do the former. Thank you!

tute:
@codyrobbins, are you still interested in making this change? Thank you!

codyrobbins:
@tute Sorry, this got lost in my notifications—I just opened a pull request with the change.

tute:
Closing in favor of PR https://github.com/thoughtbot/paperclip/pull/2279. Thank you!

PR #2279, opened 2016-08-19
Updated documentation

codyrobbins:
[These lines](https://github.com/thoughtbot/paperclip/blob/95acf3b898bd782f4429d42fd2d8812505390328/lib/paperclip/storage/s3.rb#L101-L112) in the documentation indicated that the value for the `s3_storage_class` configuration option should be `:reduced_redundancy`. However, using this value with Paperclip 5.0.0 results in a `Aws::S3::Errors::InvalidStorageClass` error with the message `The storage class you specified is not valid`.

The [documentation for the `Aws::S3::Object#storage_class` method](http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html#storage_class-instance_method) in the `aws-sdk` gem documentation indicates that the correct value is now `REDUCED_REDUNDANCY` (along with a few other potential values).

### DIFF
--- a/lib/paperclip/storage/s3.rb
+++ b/lib/paperclip/storage/s3.rb
@@ -99,17 +99,21 @@ module Paperclip
     #   "x-amz-meta-" before sending it as a header on the object
     #   upload request. Can be defined both globally and within a style-specific hash.
     # * +s3_storage_class+: If this option is set to
-    #   <tt>:reduced_redundancy</tt>, the object will be stored using Reduced
-    #   Redundancy Storage.  RRS enables customers to reduce their
+    #   <tt>:REDUCED_REDUNDANCY</tt>, the object will be stored using Reduced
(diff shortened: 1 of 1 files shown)


tute:
Thank you! Looking good.

* https://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html#storage_class-instance_method says the value should be a string. Does it work with symbols too?
* Can you please add to the commit message your good https://github.com/thoughtbot/paperclip/issues/2271#issue-169045522?

Thanks again!

codyrobbins:
Yes, I’ve been using it with symbols with no apparent problems. Sure, I just updated the comment.

tute:
I meant the commit message, but I squashed it and added it there too. Thanks a lot! 🎉 😃

codyrobbins:
Oops, sorry about that! And no problem!

Issue #2071, opened 2015-12-18
S3 storage path option still defaults to Rails.root/...


joallard:
Using `has_attached_file :column, storage: :s3, url: "..."`, the complete path will still contain Rails.root, meaning in development I get an URL like `bucket.s3.amazonaws.com/users/jon/home/dev/project/public/blah/foo/bar` which doesn't really make sense.

Moreover, adding `path: ''` will completely nix the path to the file.

Feature request: Have a sensible default work out-of-the-box.

joallard:
Okay, looks like I have to use the `path` option, not `url`. Looks like it's a documentation issue. 

[`#has_attached_file`](http://www.rubydoc.info/gems/paperclip/Paperclip/ClassMethods#has_attached_file-instance_method) mentions `url` right away, so testing on the local filestorage before switching to s3 sets that expectation. No mention of the `path` parameter there.

tute:
Thanks for pointing this out, @joallard. Would you please submit a fix please?

duhaime:
@joallard Are you pursuing this? I'm also interested in fixing this, and would be happy to help but didn't want to duplicate effort...

joallard:
@duhaime Nope, I used a workaround and the project was short-lived, so it's not really important for me anymore. Go nuts!

mike-burns:
Thank you for reporting this. Unfortunately, we will be deprecating Paperclip and therefore will not have the bandwidth to address this issue.

PR #2282, opened 2016-08-19
Documentation changes

andyatkinson:
See #2071

### DIFF
--- a/lib/paperclip.rb
+++ b/lib/paperclip.rb
@@ -129,6 +129,8 @@ module Paperclip
     #   Paperclip::Attachment#interpolate for more information on variable interpolaton.
     #     :url => "/:class/:attachment/:id/:style_:filename"
     #     :url => "http://some.other.host/stuff/:class/:id_:extension"
+    # Note: When using the +s3+ storage option, the +url+ option expects particular values.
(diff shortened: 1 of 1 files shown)


andyatkinson:
Summarizing the two issues, I came up with: 

If a person uses "path" and "url" at the same time, it produces a weird result.

If a person uses "path" and the "s3" option, it produces a weird result.

Not sure the latter one was addressed here. Will poke around a little more.

lib/paperclip.rb
```diff
@@ -129,6 +129,8 @@ module ClassMethods
     #   Paperclip::Attachment#interpolate for more information on variable interpolaton.
     #     :url => "/:class/:attachment/:id/:style_:filename"
     #     :url => "http://some.other.host/stuff/:class/:id_:extension"
+    # Note: When using the +s3+ storage option, the +url+ option expects particular values.
+    # See the Paperclip::Storage::S3#url documentation for specifics.
```
tute: These two lines should be intended two spaces if they are meant to extend on the previous point.

tute:
Merged your changes into master. We can continue working through it. Thank you!

PR #2284, opened 2016-08-23
Ignore invalid Content-Disposition header

### LINKED ISSUE #2283: NoMethodFound exception on invalid Content-Disposition header
Tried to upload file by URL

The server returned 
``"content-disposition"=>"inline; filename=blabla.jpg"``
filename must be quoted so the header is invalid
regexp ``match(/filename="([^"]*)"/)`` finds nothing here so ``match(...)[1]`` fails

I think that it's good to just ignore it instead of trying to parse it anyway

sandfoxme:
Here is one way to fix #2283

### DIFF
--- a/lib/paperclip/io_adapters/uri_adapter.rb
+++ b/lib/paperclip/io_adapters/uri_adapter.rb
@@ -29,7 +29,8 @@ module Paperclip
 
     def filename_from_content_disposition
       if @content.meta.has_key?("content-disposition")
-        matches = @content.meta["content-disposition"].match(/filename="([^"]*)"/)
+        matches = @content.meta["content-disposition"].
+          match(/filename="([^"]*)"/)
         matches[1] if matches
       end
     end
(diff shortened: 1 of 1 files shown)


tute:
There is no behavior change now. Does this fix #2283?

sandfoxme:
Yes, the exception is now gone

tute:
Thank you!

maletor:
Is it possible we are almost ready to cut a new release with this regression fix?

jbielick:
Is there a chance this could be backported to 4.3?

phbernard:
As an aside, [a Content-Disposition is allowed to not contain a filename component](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition). So strictly speaking this fix is not only for "invalid Content-Disposition", making it even more useful.

I backup @maletor, it would be great to have a new release with this fix.

PR #2287, opened 2016-08-24
Support s3 host alias with prefixes

cjcaj:
This allows s3_host_alias to be set to CDNs which support S3 Bucket prefix


tute:
Thanks! Can you please rebase against latest master, and address Hound comments? Thanks!

tute:
Merged into master. Thank you!

PR #2289, opened 2016-08-24
Fixed only_process option with proc

morgoth:
Bug introduced in 1c7d7f66ceea27cfe604e3a13c2eb87df217d13a

Per https://github.com/thoughtbot/paperclip/commit/1c7d7f66ceea27cfe604e3a13c2eb87df217d13a#commitcomment-18765681 request /cc @tute

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -240,7 +240,8 @@ module Paperclip
     # the instance's errors and returns false, cancelling the save.
     def save
       flush_deletes unless @options[:keep_old_files]
-      if @options[:only_process].any? && !@options[:only_process].include?(:original)
+      process = only_process
+      if process.any? && !process.include?(:original)
         @queued_for_write.except!(:original)
       end
       flush_writes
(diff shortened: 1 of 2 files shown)


lib/paperclip/attachment.rb
```diff
@@ -240,7 +240,8 @@ def dirty?
     # the instance's errors and returns false, cancelling the save.
     def save
       flush_deletes unless @options[:keep_old_files]
-      if @options[:only_process].any? && !@options[:only_process].include?(:original)
+      process = only_process
```
morgoth: Maybe it would be better to memoize `only_process` method?
If it's called somewhere else, proc will be computed again. What do you think?

tute: > Maybe it would be better to memoize only_process method?
> If it's called somewhere else, proc will be computed again. What do you think?

Could that present regressions? For now, I'd rather not cache it, and wait for it to actually become a bottleneck to discuss if we should go down that road. Thanks!

tute:
Thanks so much! 🎉

jaybloke:
Regardless of the contents of only_process, shouldn't the original file always be written to the file system?

I recently raised an issue when using the latest version of Paperclip with Delayed Paperclip - https://github.com/jrgifford/delayed_paperclip/issues/199

PR #2291, opened 2016-08-27
Added `:use_accelerate_endpoint` option when using S3

gagoit:
[Amazon S3 Transfer Acceleration](http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html) enables fast, easy, and secure transfers of files over long distances between your client and an S3 bucket. And [aws-sdk](https://github.com/aws/aws-sdk-ruby) has allowed us to enable this feature in version 2.3.0 (more detail is [here](https://github.com/aws/aws-sdk-ruby/pull/1163)).

You can now pass a configuration option (:use_accelerate_endpoint) when config Paperclip with S3. You can enable the [Amazon S3 Transfer Acceleration](http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html) by set this option:
```:use_accelerate_endpoint => true```

### DIFF
--- a/lib/paperclip/storage/s3.rb
+++ b/lib/paperclip/storage/s3.rb
@@ -155,7 +155,12 @@ module Paperclip
 
           @http_proxy = @options[:http_proxy] || nil
 
-          @use_accelerate_endpoint = @options[:use_accelerate_endpoint] || false
+          if @options.has_key?(:use_accelerate_endpoint) &&
+             Gem::Version.new(Aws::VERSION) < Gem::Version.new("2.3.0")
+            raise LoadError, ":use_accelerate_endpoint is only available from aws-sdk version 2.3.0. Please upgrade aws-sdk to a newer version."
+          end
+
+          @use_accelerate_endpoint = @options[:use_accelerate_endpoint]
         end
 
         Paperclip.interpolates(:s3_alias_url) do |attachment, style|
@@ -264,7 +269,7 @@ module Paperclip
       end
 
       def use_accelerate_endpoint?
-        @use_accelerate_endpoint
+        !!@use_accelerate_endpoint
       end
 
       def using_http_proxy?
--- a/lib/paperclip/storage/s3.rb
+++ b/lib/paperclip/storage/s3.rb
@@ -102,6 +102,8 @@ module Paperclip
     #   Redundancy Storage. RRS enables customers to reduce their
     #   costs by storing non-critical, reproducible data at lower
     #   levels of redundancy than Amazon S3's standard storage.
+    # * +use_accelerate_endpoint+: Use accelerate endpoint
+    #   http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html
     #
     #   You can set storage class on a per style bases by doing the following:
     #     :s3_storage_class => {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by gagoit** [comment]
Because the ```:use_accelerate_endpoint``` config is only available from [aws-sdk 2.3.0](https://github.com/aws/aws-sdk-ruby/releases/tag/v2.3.0), so The Travis CI build failed with my new tests (because it uses old version of aws-sdk). I am not sure what should I do now. Can somebody help me?

**Comment by tute** [comment]
Can we add a condition such that that behavior is triggered both if the `:use_accelerate_endpoint` option is in use, and if the right version of the SDK is available? Also, if it's not the right version, we can print out a warning, or an error message.
Thanks!

File: spec/paperclip/storage/s3_spec.rb [hunk 1]
```diff
@@ -282,6 +282,53 @@ class << @dummy
     end
   end
 
+  context "use_accelerate_endpoint" do
+    context "defaults to false" do
+      before do
+        rebuild_model storage: :s3,
+          s3_credentials: {},
```
tute: This is what Hound wants here:

```ruby
rebuild_model(
  storage: :s3,
  s3_credentials: {},
  bucket: "bucket",
  # ...
)
```

lib/paperclip/storage/s3.rb
```diff
@@ -152,6 +154,8 @@ def self.extended base
           @options[:url] = @options[:url].inspect if @options[:url].is_a?(Symbol)
 
           @http_proxy = @options[:http_proxy] || nil
+
+          @use_accelerate_endpoint = @options[:use_accelerate_endpoint] || false
```
tute: ` || false` is the same as nothing (falsy). We can drop that.

gagoit: Thank you. I updated :)

paperclip.gemspec
```diff
@@ -35,7 +35,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency('rspec', '~> 3.0')
   s.add_development_dependency('appraisal')
   s.add_development_dependency('mocha')
-  s.add_development_dependency('aws-sdk', '>= 2.0.34', '< 3.0')
+  s.add_development_dependency('aws-sdk', '>= 2.3.0', '< 3.0')
```
tute: Is this bump necessary? I wouldn't require it to everyone; the exception takes care of it for us. CI should anyway install latest version, as that's what we are specifying here.

lib/paperclip/storage/s3.rb
```diff
@@ -232,6 +241,8 @@ def s3_interface
             config[:proxy_uri] = URI::HTTP.build(proxy_opts)
           end
 
+          config[:use_accelerate_endpoint] = true if use_accelerate_endpoint?
```
tute: `config[:use_accelerate_endpoint] = true if use_accelerate_endpoint?`

is the same as

`config[:use_accelerate_endpoint] = use_accelerate_endpoint?`

tute:
Merged it in. Thank you! 🎉

gagoit:
Thanks for you updating. It's great :)

PR #2310, opened 2016-09-22
Update docs for s3_permissions

HormyAJP:
s3_permissions can no longer be a symbol. Must be a string.
Providing a string causes hard to debug errors in AWS.

As a side note, when these sorts of breaking changes occur to 
config I really think checks need to be put in place to validate
user input. Debugging this was time consuming as the errors
I was getting in no way made it obvious this was the problem.

N.B. I'm assuming that a change was made at some point to 
change s3_permissions from a symbol to a string. We were
previously using an older version of paperclip where this 
worked fine but when I updated this started failing.

### DIFF
--- a/lib/paperclip/storage/s3.rb
+++ b/lib/paperclip/storage/s3.rb
@@ -45,10 +45,10 @@ module Paperclip
     #
     #   You can set permission on a per style bases by doing the following:
     #     :s3_permissions => {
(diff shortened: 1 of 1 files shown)


tute:
> I really think checks need to be put in place to validate
> user input.

Totally agree.

Thanks for your change! :)

PR #2322, opened 2016-10-13
uri adapter: allow filenames with unquoted syntax

maletor:
There is sort of this gray area where two RFCs conflict about the format
for the filename parameter. It is legal for servers to send unquoted
filenames, though no server should do this since it complicates things.
Nevertheless, client implementations should be able to handle both kinds
of syntax.

https://tools.ietf.org/html/rfc2616
https://tools.ietf.org/html/rfc6266#section-5
http://rubular.com/r/6m0tqSfenR


mike-burns:
Fixed in fa37ab74aae280c0b00a9846b7d4920ac5a71b6c and 3155cd7ec1942cde6ec3ac0d6745c3eaee91afdd. Thank you for fixing this, too!

PR #2325, opened 2016-10-21
Check for spoofing of files without an extension

walterg2:
While using the Paperclip gem, we noticed during some ad-hoc testing that if you do not supply an extension when uploading a file, Paperclip effectively skipped it's spoofing check, which allowed potentially dangerous files to slip through into your application.

This addresses that by moving the checks around a little bit and only testing against the extension when there is one.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -18,6 +18,7 @@ master:
   `CLASS=User rake paperclip:refresh:fingerprints`
   You can optionally limit the attachment that will be processed, e.g:
   `CLASS=User ATTACHMENT=avatar rake paperclip:refresh:fingerprints`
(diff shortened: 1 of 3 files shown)


walterg2:
It looks like the Travis-CI builds are failing for Ruby 2.2 and both Rails 4 and 5 due to a Gherkin Lexer error. I've run these tests locally and they pass, but I would understand if you're not comfortable in passing this PR with those failures. Just let me know how I can help resolve these errors.

mike-burns:
Merged as 7eb664f21ba1cbcaf2510b46055878241eb8c5b2. Thank you!

PR #2331, opened 2016-10-28
Fix Travis CI

tute:
- Removes `bundle exec` from commands ran in cucumber steps
- Simplifies `framework_major_version` (we can assume it always is 4 or higher)

### DIFF
--- a/features/support/rails.rb
+++ b/features/support/rails.rb
@@ -35,29 +35,5 @@ module RailsCommandHelpers
   def framework_major_version
     framework_version.split(".").first.to_i
   end
-
-  def using_protected_attributes?
-    framework_major_version < 4
-  end
-
-  def new_application_command
-    "rails new"
-  end
-
-  def generator_command
-    if framework_major_version >= 4
-      "rails generate"
-    else
-      "script/rails generate"
-    end
-  end
-
-  def runner_command
-    if framework_major_version >= 4
-      "rails runner"
-    else
-      "script/rails runner"
-    end
-  end
 end
 World(RailsCommandHelpers)
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ rvm:
 
 script: "bundle exec rake clean spec cucumber"
 
+before_install:
+  - rvm @global do gem uninstall bundler -a -x
+  - rvm @global do gem install bundler -v 1.12.5
+
 gemfile:
   - gemfiles/4.2.gemfile
   - gemfiles/5.0.gemfile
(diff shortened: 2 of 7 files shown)


tute:
Calling in @sikachu for a review of this PR. Thank you! :)

features/support/rails.rb
```diff
@@ -36,28 +36,16 @@ def framework_major_version
...
+    "rails generate"
   end
 
   def runner_command
-    if framework_major_version >= 4
-      "rails runner"
-    else
-      "script/rails runner"
-    end
+    "rails runner"
```
tute: Should we just inline these calls?

tute:
@sikachu I do not know why this started failing, would love to know.

features/support/rails.rb
sikachu: Are we dropping Rails 3? If so, then definitely.

tute: We did, yes!

tute:
Maybe related: http://stackoverflow.com/questions/15123677/how-do-i-resolve-a-cucumber-cant-load-such-file-warning

tute:
Maybe related: https://github.com/cucumber/gherkin2/issues/320#issuecomment-139572490

tute:
`be rake cucumber` fails, `be cucumber` goes through in Ruby 2. Seems related to our `Rakefile`.

tute:
`be rake cucumber` fails, `be cucumber` goes through in Ruby 2. `rake cucumber` seems to run:

`/Users/tutec/.rbenv/versions/2.2.2/bin/ruby -S bundle exec cucumber --format progress`

`% ruby -v # => ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin15]`

If I run `ruby -S bundle exec cucumber --format progress`, it works well.

I don't understand where the code loading issue is coming from.

tute:
See:

- https://github.com/thoughtbot/scenic/blob/master/spec/acceptance_helper.rb
- https://github.com/thoughtbot/scenic/blob/master/spec/acceptance_helper.rb

brendon:
Hi @tute, you can get rid of the deprecation warning regarding raise in transactional callbacks via this in `spec_helper.rb`:

```ruby
if ActiveRecord::VERSION::STRING >= '4.2' &&
ActiveRecord::VERSION::STRING < '5.0'
  ActiveRecord::Base.raise_in_transactional_callbacks = true
end
```

4.2 required the option to be set to true to prevent the deprecation warning, and 5.0+ doesn't have the option and emits another warning if it's present.

PR #2332, opened 2016-10-31
fix ImageMagick command in Windows

aboutqx:
Add OS detect to choose right command.

It seems that paperclip use identify as exe in windows,

`Command :: SET PATH=C:\file-5.03-bin\bin;%PATH% & identify -format '%wx%h,%[exif:orientation]' "C:/Users/AppData/Local/Temp/5415c7fe534a63366b48d388f204c32820161028-7156-f2nblm.jpg[0]" 2>NUL`.,

while only `magick identify` command  works in windows,and there in no identify.exe in windows in the latest version ImageMagic.
And only `Paperclip.options[:command_path]` is used, the `Paperclip.options[:image_magick_path]` seems ignored.

About imageMagick in windows can be found [there](http://www.imagemagick.org/script/binary-releases.php).

### DIFF
--- a/lib/paperclip.rb
+++ b/lib/paperclip.rb
@@ -98,7 +98,8 @@ module Paperclip
       :swallow_stderr => true,
       :content_type_mappings => {},
       :use_exif_orientation => true,
-      :read_timeout => nil
+      :read_timeout => nil,
+      :is_windows => Gem.win_platform?
     }
   end
 
--- a/lib/paperclip/geometry_detector_factory.rb
+++ b/lib/paperclip/geometry_detector_factory.rb
@@ -17,7 +17,7 @@ module Paperclip
         orientation = Paperclip.options[:use_exif_orientation] ?
           "%[exif:orientation]" : "1"
         Paperclip.run(
-          Gem.win_platform?"magick identify":"identify",
+          (Gem.win_platform?)?"magick identify":"identify",
           "-format '%wx%h,#{orientation}' :file", {
             :file => "#{path}[0]"
           }, {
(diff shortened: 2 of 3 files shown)


tute:
Thank you, @aboutqx! I like this change.

I know master is now failing in CI for Ruby 2.2, I'm taking care of that. That makes it hard to do what I'll ask, but do you know you can set up paperclip for Windows in Travis CI with a regression test for this? Otherwise, I can't make sure we can reliably maintain this code.

Also, do you think we can abstract out this change to a single place in the codebase, and then call that configuration instead of checking three times whether we are on windows or not?

Thanks so much!

tute:
> the Paperclip.options[:image_magick_path] seems ignored.

Is this a bug? If so, we should address it. Thanks!

aboutqx:
Glad for your reply.I've make it configurable by adding it to options,but I'm not familiar with Travis CI.

I've made some google,it seems that Windows is not supported in Travis CI right now,and I have no ideas.
Do you have nay solution?
Btw,the `Paperclip.options[:image_magick_path]` is never used,not sure if it's needed.

tute:
> It seems that Windows is not supported in Travis CI right now

huh, we can't test it then. We _could_ tag those tests with `windows_platform` or alike, ignored by CI, so that people on Windows platforms can run them. But I don't have available window boxes to run them myself and verify the tests and code indeed work.

> The Paperclip.options[:image_magick_path] is never used,not sure if it's needed.

Right, thanks! Just dropped the option.

aboutqx:
Maybe another way is make the `option is_windows` default false ,and add some tip about changing it in windows while referencing  the doc of ImageMagick.

Right now the error information didn't cover it,just return `not idetentified`,which is not clear.

mike-burns:
Merged as 0d93e0f0ef84d0030e803af9e3c3513c908ffa57. Thank you!

PR #2338, opened 2016-11-11
Remove Tempfile tests from Thumbnail spec

farisj:
## What's Up

Was browsing the paperclip test suite and noticed that the `thumbnail_spec` had a few tests from a while back that did not belong - they were testing `Paperclip::Tempfile` instead.

## What This Does

This PR removes those tests.

### DIFF
--- a/spec/paperclip/tempfile_spec.rb
+++ b/spec/paperclip/tempfile_spec.rb
@@ -1,4 +1,4 @@
-require 'spec_helper'
+require "spec_helper"
 
 describe Paperclip::Tempfile do
   context "A Paperclip Tempfile" do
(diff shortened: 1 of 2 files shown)


tute:
Thanks! Do you think we should move over those specs to a `spec/paperclip/tempfile_spec.rb` file?

farisj:
@tute done! The test suite is failing due to a Capybara issue - perhaps this will have to wait until your work in #2331 is merged?

Thanks for looking at my PR.

tute:
That test PR needs more work than I can put in now, and this PR is ready. Merging. Thank you! :)

PR #2342, opened 2016-11-18
Skip validity check when reprocess attachment

### LINKED ISSUE #2078: Refresh styles fails when no content type column but attribute validated
The readme indicates Paperclip doesn't need other db columns but file name to work.

Given:
```
  attr_accessor :photo_content_type
  has_attached_file :photo, 
    styles: { medium: "250x250", thumb: "100x100" }
  validates_attachment_content_type :photo,
    content_type: %w(image/jpeg image/png)
```

Issuing `rake paperclip:refresh:missing_styles` will trigger a validation error

sikachu:
This validation was causing issue when there is no content_type column in the database, as the content type of the attachment would be empty. There should be no need to check attachment's validity because this `#reprocess!` method is running internally on the server.

Fix #2078

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -337,8 +337,15 @@ module Paperclip
     # inconsistencies in timing of S3 commands. It's possible that calling
     # #reprocess! will lose data if the files are not kept.
     def reprocess!(*style_args)
-      saved_only_process, @options[:only_process] = @options[:only_process], style_args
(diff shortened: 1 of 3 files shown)


tute:
Too bad paperclip tests started failing, but this shouldn't hold off new PRs (failures are the same as in master).

This looks good to me, thanks for addressing!

mike-burns:
Whew, finally merged. Thank you again, Prem!

PR #2352, opened 2016-11-29
Fix JRuby binary operator warning

dzsibi:
Fixes the following warning shown by JRuby 9.1.6.0:

gems/paperclip-5.1.0/lib/paperclip/storage/filesystem.rb:49: warning: `&' after local variable or literal is interpreted as binary operator even though it seems like argument prefix

### DIFF
--- a/lib/paperclip/storage/filesystem.rb
+++ b/lib/paperclip/storage/filesystem.rb
@@ -46,7 +46,7 @@ module Paperclip
             end
           end
           unless @options[:override_file_permissions] == false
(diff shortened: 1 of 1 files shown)


tute:
Thanks! Can you please push against latest master, to check if CI passes? Thanks!

dzsibi:
Here ya go.

tute:
Excellent, thank you!

YavorIvanov:
I just realize that we are still using the gem from the git repo for this fix.
Should we not release 5.1.1 with this?

PR #2353, opened 2016-12-02
Fix access denied / file exists error on Windows

dzsibi:
After the changes introduced by #2290, uploads were broken for me on my Windows dev machine. The problem turned out to be the 'destination' file, which was opened and created before the 'link_or_copy_file' method is called. 'FileUtils.ln', when called with the 'force' parameter, tries to first delete the file and ignores the inevitable access denied error, then fails to create the link, since the destination path already exists. I have reordered the code to first close the destination file and only reopen it at the end, when file operations have completed.

### DIFF
--- a/lib/paperclip/io_adapters/abstract_adapter.rb
+++ b/lib/paperclip/io_adapters/abstract_adapter.rb
@@ -57,13 +57,15 @@ module Paperclip
     end
 
     def link_or_copy_file(src, dest)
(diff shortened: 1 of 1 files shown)


Pingviinituutti:
I had the same problem on my Windows machine and the changes in this pull request fixed this even though this pr is almost two years old. Maybe merge this to master or what more should be done?

sidraval:
This is already fixed on master.

PR #2360, opened 2016-12-09
Inherit toplevel Aws region configuration

aripollak:
This is already possible for the other S3 credential parameters, but
just not for region. It can be configured directly on the Aws library
with ENV['AWS_REGION'] or Aws.config[:region].


blackjid:
any plans to merge this??

Issue #1924, opened 2015-07-06
CSV upload: has contents that are not what they are reported to be


iashanmugavel:
I am getting this error while uploading csv file  in paperclip -v 4.3. For image files it is working.

Error: "has contents that are not what they are reported to be".

chetan4664:
I am getting the same error while uploading images as well. 
gem 'paperclip', '~> 4.2.1'

If I use the latest version, it goes in an infinite loop !

jyurek:
This is because it's hitting the media-type validation. This means that either the name or content type of the file being uploaded does not match the content type of the file as determined on the server. What is the name of the file you're trying to upload? Does it end in ".csv"? Are you using a custom server or hosting on heroku?

iashanmugavel:
while uploading .csv file. It is not a content type validation. same file working well with paperclip 4.2.1.

srogers:
I'm seeing the same thing with PDF attachments. Works fine with 4.2.1, but gets the "has contents that are not what they are reported to be" error in 4.3.  The error doesn't show up in dev environment where the attachment is being saved locally. It only happens in staging or production on Heroku where the attachment is being saved to S3.

bwlang:
I see this problem for CSV files...  for uploads from chrome and ie on windows
see:
https://code.google.com/p/chromium/issues/detail?id=139105
reported there that this content type comes from the windows registry...
This is such a standard case that I don't think users should have to configure special mime types...

sgringwe:
@bwlang i was able to fix a similar situation (csv files, windows) with this in the initializer:

```
comma_separated_values = MIME::Types["text/comma-separated-values"].first
comma_separated_values.extensions << "csv"
MIME::Types.index_extensions comma_separated_values
```

bwlang:
@sgringwe 
Yes - your suggestion works... 
I did this instead;  (which avoids the deprecated index_extensions method)

  Paperclip.options[:content_type_mappings] = {:csv => "text/plain"}

However, I suspect this will bite almost anyone who want to handle csv file uploads from windows.  I'm thinking about how to reasonably make this a better experience for devs.

options that occur to me:
 - extend https://github.com/mime-types/ruby-mime-types/blob/master/type-lists/application.yaml to include csv for application/vnd.ms-excel 
 - modify paperclip to stop caring about application/* - not sure yet if this is a security problem (maybe we don't care if application a attempts to parse data from billed as application b)
[...]

PR #2378, opened 2017-01-24
Added config to bypass media type spoofing check

rystraum:
https://github.com/thoughtbot/paperclip/issues/1924

Though a better implementation for this would be doing this on a per
model basis.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,16 @@
 master:
 
+* Improvement: Added `:do_not_do_media_type_spoofing_detection` option to Paperclip
+  to disable media type spoofing. Internally, this just returns `false` when calling 
+  `#spoofed?` on `MediaTypeSpoofDetector` when the configuration is set.
+
+  This is a reduction in security (hence the long and explicit config key) but this
+  would allow developers to make a decision on enabling/disabling Media Type Spoof
+  Detection on their environments.
+
+  A better implementation for this would be implementing this on a per-class basis
+  instead of a global configuration.
+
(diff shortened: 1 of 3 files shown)


jtrost:
This is great! I think `:do_not_do_media_type_spoofing_detection` is too long and verbose for a configuration key. It could be more concisely named `:detect_media_type_spoofing`.

rystraum:
@jtrost I opted for a verbose configuration key, following Paperclip's `do_not_validate_attachment_file_type` to tell the developer that this isn't a recommended setting, but you can change it if you really want.

jtrost:
Thanks for the explanation!

kirushanth-sakthivetpillai:
An alternative:

Paperclip has content-type checking and media-type checking. Both seem to do the same thing using `file -b --mime`.

Media-type checking uses the global config of `content_type_mappings`. However, sometimes you want per-model exceptions. Paperclip provides the option for it but not well documented. I got it from the source here:
https://github.com/thoughtbot/paperclip/blob/v4.3/lib/paperclip/has_attached_file.rb#L83

You can do the following to control media-type and content-type checking:

```
class PaperclipModel < ActiveRecord::Base

   has_attached_file :attachment, { validate_media_type: false }

  validates_attachment :attachment, {
     # tweak as desired
     content_type: { content_type: ["text/csv", "text/plain", Paperclip::ContentTypeDetector::SENSIBLE_DEFAULT] }
  }
end
```

saghaulor:
@rystraum Paperclip can disable media type spoof checking at a per model level with the `validate_media_type` option.

PR #2392, opened 2017-02-22
Better handling of the content-disposition header

Whyves:
Encountered an issue where the URI was returning header with content-disposition where the filename value wasn't enclosed in the double quotes. Turns out that this is a valid grammar according to RFC6266. Also made the logic more robust to account for spaces and uppercase letters.


mike-burns:
Merged as fa37ab7. Thank you!

PR #2401, opened 2017-03-02
Ruby 2.4.0 compatibility (doen't use Fixnum anymore)

ledermann:
On Ruby 2.4.0, this PR avoids warnings like this:

```
/usr/local/rvm/gems/ruby-2.4.0/gems/paperclip-5.1.0/lib/paperclip/storage/fog.rb:184: warning: constant ::Fixnum is deprecated 
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -18,6 +18,7 @@ master:
   `CLASS=User rake paperclip:refresh:fingerprints`
   You can optionally limit the attachment that will be processed, e.g:
   `CLASS=User ATTACHMENT=avatar rake paperclip:refresh:fingerprints`
+* Ruby 2.4.0 compatibility (doen't use Fixnum anymore)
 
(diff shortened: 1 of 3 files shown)


NEWS
```diff
@@ -18,6 +18,7 @@ master:
   `CLASS=User rake paperclip:refresh:fingerprints`
   You can optionally limit the attachment that will be processed, e.g:
   `CLASS=User ATTACHMENT=avatar rake paperclip:refresh:fingerprints`
+* Ruby 2.4.0 compatibility (doen't use Fixnum anymore)
```
reedloden: don't or doesn't

janko-m:
Since `Fixnum` is a descendant from `Integer`, you can use `is_a?(Integer)` instead of `is_a?(0.class)`.

Issue #2412, opened 2017-03-23
file_name is not available when setting s3_headers


cgreening:
Hi All,

I'm trying to set the content disposition header of an attachment stored on S3

```rb
  has_attached_file :attachment, :storage => :s3, :s3_protocol => 'https', ...
                    :s3_headers => lambda { |asset|
                      { "Content-Disposition" => "attachment; filename=\"#{asset.attachment_file_name}\"" }
                    }
```

Unfortunately attachment_file_name doesn't seem to to be set when the lambda is called so I end up with `attachment; filename=""` in the header.

Is there any way to get the original filename in the s3_headers lambda?

Thanks

simonfranzen:
I am having the same problem. I need to set here csv or xlsx format.

Saving CSV to S3 without explicitly setting Content-Type and Content-Disposition will save the file in wrong content type. I also tried it to merge the extras in a post_process callback, but this also had no effect.

mike-burns:
Thank you for the feature request. Unfortunately we will be deprecating Paperclip and therefore will no longer be accepting new features.

zainzafar:
@simonfranzen what solution did you end up with?

PR #2413, opened 2017-03-27
Adds optional style and file argument to s3_headers block

cgreening:
Fixes https://github.com/thoughtbot/paperclip/issues/2412

There are now two additional parameters that get passed to the s3_headers block:

```rb
s3_headers: lambda {|attachment, style, file|
  {'Content-Disposition' => "attachment; filename=\"#{file.original_filename}\""}
}
```

### DIFF
--- a/lib/paperclip/storage/s3.rb
+++ b/lib/paperclip/storage/s3.rb
@@ -458,7 +458,7 @@ module Paperclip
         # keep backwards compatible with previous version that only took one argument
         http_headers = http_headers.call(instance) if http_headers.respond_to?(:call) && http_headers.arity == 1
         http_headers = http_headers.call(instance, style, file) if http_headers.respond_to?(:call) && http_headers.arity == 3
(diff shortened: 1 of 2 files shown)


mike-burns:

bterkuile:
This seems an interesting addition indicating one of the issues I had to write a work-around for as wel to use the file's `original_filename` in the S3 `Content-Disposition` data. One of the problems however still seems to be that the `s3_headers` lambda is evaluated at initialization time and not on file assignment (`send(name).assign(file)`) moment so probably in the example:
```ruby
record  = MyModel.new(file: file1)
record.file = file2
record.save
```
will set the `original_filename` of `file1`. If you know that a model's file will only be set once not a big problem. Then also consider returning an empty hash `{}` for persisted record since if for dynamically generated filenames with for example a timestamp in it the `Content-Disposition` header will not reflect the file name (probably minor consistency issue).

Lets hope that the ActiveStorage will be able to handle these cases nicely.

PR #2415, opened 2017-03-29
Include Cocaine error when there was an error processing thumbnails

ball-hayden:
We have been occasionally seeing errors in production telling us "There was an error processing the thumbnail". Unfortunately, Paperclip doesn't currently tell us _what_ was ImageMagick was complaining about, so we weren't able to debug further what was going on.

This PR ensures that the original error message from Cocaine (and therefore ImageMagick) is included in the Paperclip error (when `whiny` mode is on), which allowed us to identify the problem with our ImageMagick configuration.


spec/paperclip/thumbnail_spec.rb
```diff
@@ -211,10 +211,10 @@
         end
 
         it "errors when trying to create the thumbnail" do
-          assert_raises(Paperclip::Error) do
-            silence_stream(STDERR) do
+          silence_stream(STDERR) do
+            expect {
```
ball-hayden: Hmm. A continual conundrum for me.

```ruby
expect do
  blah
end.to raise_error
```

I think I prefer `{...}`, but what's your take?

lib/paperclip/thumbnail.rb
```diff
@@ -78,7 +78,10 @@ def make
 
         success = convert(parameters, :source => "#{File.expand_path(src.path)}#{'[0]' unless animated?}", :dest => File.expand_path(dst.path))
       rescue Cocaine::ExitStatusError => e
-        raise Paperclip::Error, "There was an error processing the thumbnail for #{@basename}" if @whiny
+        if @whiny
+          message = "There was an error processing the thumbnail for #{@basename}:\n" + e.message
```
ball-hayden: I think I prefer a long line here, but I'm happy to split it.

Issue #2422, opened 2017-04-08
Allow to choose frame (page) from a PDF file


adaRn:
This was already reported in #2155 and almost fixed.
Probably due to an oversight `lib/paperclip/thumbnail.rb` is still missing `.pdf` in `MULTI_FRAME_FORMATS` array. 

Another problem that arises is that if no frame (page) is chosen, current implementation falls back on `[0]` frame. But there are cases when frame mustn't be declared explicitly. For example no frame declaration in conjunction with `-append` parameter will make `convert` yield one image in which the pages are stacked vertically.

adaRn:
I made a PR addressing the first part of the issue. I have no idea how to come at the second problem. I've tried setting `frame_index` to `''` or `*`, meaning `convert` would get `[]` or `[*]`, but none of them work.

mike-burns:
Thank you for the feature request. Unfortunately we will be deprecating Paperclip and therefore will no longer be accepting new features.

PR #2423, opened 2017-04-08
Allow user to choose a page from PDF file 

adaRn:
Add .pdf to MULTIFRAME_FORMATS [adresses #2422]


mike-burns:

PR #2429, opened 2017-04-17
Private filesystem attachments

nextgenappsllc:
I came across multiple instances where I need to protect the file downloads from the public or even from certain users. 

For instance when a form gets submitted I want only the person that submits the form or the person the form is destined for to be able to download attachments for that form.

Instead of creating a concern or an action in each controller for downloads I wanted to define the attachment as private and have it all routed through one controller.

I would also like some aid in perfecting the changes I made.

**I added the following to the readme to explain some of the changes:**

---

Private Attachments
--------------------------

If you want to place files behind a controller in order to perform validation you can.

In your routes.rb file mount the paperclip engine with any path you'd like. 
Ex:
```ruby
mount Paperclip::Engine => 'paperclip/'
```

In your ApplicationController add the method `paperclip_whitelist` which returns an array of the models with attachments you would like to serve privately. 
Ex:
```ruby
def paperclip_whitelist
  [PrivateAttachment]
end
```

Then in your model add `privacy: :private` to the has_attached_file options. Then add the method `can_download_attachment?`. 
The method `can_download_attachment?` gets passed the controller instance and the params and is expected to return true, false, or raise Paperclip::Errors::AccessDeniedError.
The controller is passed so that methods like current_user can be run on it to get the user instance for validation puprposes. Duplicated params gets passed to do things like allow anyone if the style is :thumb or to only allow :original to paid users.
Ex:
```ruby
  has_attached_file :file, privacy: :private

  def can_download_attachment?(controller, params)
    params[:style] == :thumb || (!controller.current_user.nil? && created_by == controller.current_user.id)
  end
```


nextgenappsllc:
I'll write the tests dependent on if the feedback is good. If not I will migrate this code out into its own gem.

nextgenappsllc:
Alright I'll guess I'll migrate this code into an external gem

mike-burns:

PR #2438, opened 2017-04-29
Fix S3 HTTP Proxy option passed to AWS SDK, which changed in version 2

douglaslise:
In version 2, aws-sdk started using Seahorse as HTTP Client, and the
`http_proxy` configuration changed from `:proxy_uri` to `:http_proxy`.

The new option can be seen here:
https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/lib/seahorse/client/net_http/connection_pool.rb#L234


mike-burns:
We have switched to aws-sdk-s3.

PR #2447, opened 2017-05-14
Add Rails Version to migration files. Required in Rails 5.1+

Arinzeokeke:


sidraval:
Closing since a very similar change made it into master

Issue #2465, opened 2017-07-09
Directly inheriting from ActiveRecord::Migration is not supported


acandael:
this error is given when trying to run a migration generated with the paperclip generator:

    rails g paperclip absence document

the error is:

    Directly inheriting from ActiveRecord::Migration is not supported. Please specify
     the Rails release the migration was written for:

the Rails version I use is 5.1.2

I fixed the issue by putting
    [5.1] 
at the end of ActiveRecord::Migration:

    class AddAttachmentDocumentToAbsences < ActiveRecord::Migration[5.1]

jackellenberger:
Same issue, same solution. 
```
diff --git a/db/migrate/20170718022803_add_attachment_image_to_posts.rb b/db/migrate/20170718022803_add_attachment_image_to_posts.rb
index accf44e..93f1785 100644
--- a/db/migrate/20170718022803_add_attachment_image_to_posts.rb
+++ b/db/migrate/20170718022803_add_attachment_image_to_posts.rb
@@ -1,4 +1,4 @@
-class AddAttachmentImageToPosts < ActiveRecord::Migration
+class AddAttachmentImageToPosts < ActiveRecord::Migration[5.1]
   def self.up
     change_table :posts do |t|
       t.attachment :image
```

stinnett-casey:
Same issue, same work around. I think a better solution would be to have Paperclip get the current Rails version and add that automatically. Unless I am missing something?

PR #2470, opened 2017-07-18
Add migration_version variable to migration generator for rails 5 compatibility

jackellenberger:
# Problem

[Relevant Issue](https://github.com/thoughtbot/paperclip/issues/2465) : Rails 5 does not allow classes to inherit directly from ActiveRecord::Migration without specifying a version. This prevents basic actions such as `db:migrate`.

# Solution

If the rails version is >= 5.x, pass a `migration_version` variable to the `migration_template` function to ensure that `[5.x]` is appended to the migration function declaration.

For similar code from a more trusted source, see [Devise's implementation](https://github.com/plataformatec/devise/blob/master/lib/generators/active_record/devise_generator.rb#L14).

## Concerns

Hey there's no test for this change! But there are no tests for any of the paperclip generator functions. It's late maybe I'll write em later.

### DIFF
--- a/lib/generators/paperclip/paperclip_generator.rb
+++ b/lib/generators/paperclip/paperclip_generator.rb
@@ -13,7 +13,7 @@ class PaperclipGenerator < ActiveRecord::Generators::Base
   end
 
   def generate_migration
-    migration_template "paperclip_migration.rb.erb", "db/migrate/#{migration_file_name}"
+    migration_template "paperclip_migration.rb.erb", "db/migrate/#{migration_file_name}", migration_version: migration_version
   end
 
   def migration_name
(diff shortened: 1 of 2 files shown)


stinnett-casey:
This looks like a great solution. What is stopping this from being merged?

lib/generators/paperclip/paperclip_generator.rb
```diff
@@ -27,4 +29,10 @@ def migration_file_name
   def migration_class_name
     migration_name.camelize
   end
+
+  def migration_version
+    if Rails.version.start_with? "5"
```
Fernan2: [SUG] Let's do it compatible with higher versions:
```
if Rails.version.to_i >= 5
```

christopherie: This worked perfectly.

Skowt:
Any reason this still hasn't been merged 🤔

PR #2476, opened 2017-08-02
Remove bad exception handling!

gaurish:
Exception is the root of Ruby's exception hierarchy, so when you rescue Exception you rescue from everything, including subclasses such as SyntaxError, LoadError, and Interrupt etc which is a bad idea.

Fix: rescue from StandardError.

### DIFF
--- a/lib/tasks/paperclip.rake
+++ b/lib/tasks/paperclip.rake
@@ -46,7 +46,7 @@ namespace :paperclip do
           attachment = instance.send(name)
           begin
             attachment.reprocess!(*styles)
-          rescue Exception => e
+          rescue => e
             Paperclip::Task.log_error("exception while processing #{klass} ID #{instance.id}:")
             Paperclip::Task.log_error(" " + e.message + "\n")
           end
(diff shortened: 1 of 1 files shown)


lib/tasks/paperclip.rake
```diff
@@ -46,7 +46,7 @@ namespace :paperclip do
           attachment = instance.send(name)
           begin
             attachment.reprocess!(*styles)
-          rescue Exception => e
+          rescue => e
```
pjg: I think it would be even better to explicitly say `rescue StandardError => e`

geoffharcourt: Agreed, this appears to have the same problem as before, it's just now even broader. Agree in principle that rescuing from `StandardError` would be an improvement unless there's some system reason that we needed to use `Exception` here. Since it's not covered by a test I think we can assume this was just an oversight.

gaurish:
Updated. Ready for another pass.

geoffharcourt:
Looks great! Thank you for your contribution.

PR #2478, opened 2017-08-23
Backport Invalid Content-Disposition header fix to v4.3 

jbielick:
This cherry picks 1f7886b onto the `v4.3` branch and tags a release for `v4.3.8`—backporting an important fix for the `UriAdapter` so that files can be attached with or without a valid `Content-Disposition` header. In the invalid or missing case, an error occurs as described in #2283.

#2284 Added this to the `master` branch and this adds that commit to the 4.3 history.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,7 @@
+4.3.8 (8/23/2017):
+
+* Bug Fix: Backport "Ignore invalid content-disposition header (#2284)" to 4.3
+
 4.3.7 (7/1/2016):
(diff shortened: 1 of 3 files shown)


mike-burns:
We are in the process of deprecating paperclip, and so are closing pull requests that aren't immediate bug fixes or documentation updates. This backport is something we will not have the bandwidth to handle. Thank you for your contribution, we regret not tending to it!

PR #2488, opened 2017-09-17
Fix build by using precise distro, added more Ruby versions to test

morgoth:


morgoth:
@geoffharcourt Looks like there was a change in default travis image which after, paperclip tests are failing. https://blog.travis-ci.com/2017-08-29-trusty-image-updates
I tried to use previous one, but it fails as well.
I can try to write Travis support unless you have some other findings or want to do it on your own.

josephpage:
Hey @morgoth, I found `For the sudo: false images, this workaround is not available.` in the blog post.
Also, why use `dist: trusty` and not `language: ruby` ?

morgoth:
@josephpage Tried different configurations ;-) No luck so far

morgoth:
@josephpage Thanks for looking into it. Unfortunately it still fails https://travis-ci.org/thoughtbot/paperclip/builds/280410099?utm_source=github_status&utm_medium=notification

josephpage:
Found !
The `ghostscript` dependencies is missing (which depends on others : `libcupsfilters1 libcupsimage2 libgs9 libgs9-common libijs-0.35 libjbig2dec0 libpaper-utils libpaper1 poppler-data`)

```
addons:
  apt:
    packages:
    - ghostscript
```
Successfully tested locally.

josephpage:
Yeah 🎉  the errors are not the same !

morgoth:
I'm getting totally different failure locally, but also with cucumber...
```
    Given I generate a new rails application                                            # features/step_definitions/rails_steps.rb:1
    And I run a rails generator to generate a "User" scaffold with "name:string"        # features/step_definitions/rails_steps.rb:82
      expected "rails generate scaffold User name:string" to be successfully executed (RSpec::Expectations::ExpectationNotMetError)
      ./features/step_definitions/rails_steps.rb:83:in `/^I run a rails generator to generate a "([^"]*)" scaffold with "([^"]*)"$/'
      features/rake_tasks.feature:5:in `And I run a rails generator to generate a "User" scaffold with "name:string"'
```

josephpage:
I'm getting exactly the same !! 😞

josephpage:
To reproduce the same errors as in Travis builds, you need to execute `export BUNDLE_GEMFILE=$PWD/gemfiles/5.0.gemfile` before `bundle install`
See : https://travis-ci.org/thoughtbot/paperclip/jobs/280457277#L575

.travis.yml
```diff
@@ -1,13 +1,25 @@
...
 rvm:
   - 2.1
-  - 2.2.2
-
-script: "bundle exec rake clean spec cucumber"
+  - 2.2
+  - 2.3
+  - 2.4
 
+# FIXME: fails with modern bundler
```
josephpage: Not sure modern bundler is the problem. I have locally tested with bundler 1.12.5 and I got the same errors.

morgoth:
Now I'm confused.
It now passes, but it says it's still using latest bundler `Using bundler 1.15.4` during gems install.
Anyway, it's green - I'm not touching it :-)

@josephpage Thank you for your help.

@geoffharcourt Looks like this PR fixes master build - can you have a look please?

josephpage:
Thank you very much @morgoth ! 🎉 

2 notes : 
- maybe we should explicitly mention `imagemagick` in the addons.apt.packages key, to prevent future issues with dependencies. As `ghostscript` is a [optional dependency of imagemagick (packages.ubuntu.com)](https://packages.ubuntu.com/trusty/imagemagick), we should mention both (maybe in the README.md too).
- I confirm this locally works with modern bundler (1.15.4)...

geoffharcourt:
@morgoth thank you very much for fixing these build problems!

@josephpage I noticed when I was testing locally that I hadn't installed imagemagick on this particular machine and had some issues. If you can do a PR to update dependency checking and/or documentation I'd be happy to review.

PR #2493, opened 2017-10-07
Fix a build

morgoth:
let's see what Travis says

### DIFF
--- a/features/step_definitions/rails_steps.rb
+++ b/features/step_definitions/rails_steps.rb
@@ -1,6 +1,6 @@
 Given /^I generate a new rails application$/ do
   steps %{
-    When I run `rails new #{APP_NAME} --skip-bundle`
+    When I successfully run `rails new #{APP_NAME} --skip-bundle`
     And I cd to "#{APP_NAME}"
   }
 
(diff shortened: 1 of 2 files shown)


morgoth:
@geoffharcourt This fixes the build again. I didn't dig enough to find a real cause - debugging cucumber is painful... ;-)
Looking at recent failures, I would be for checking in `Gemfile.lock` into repo, to avoid such situations, as it slows down development a lot.

It's also now the recommended workflow by bundler, vide https://github.com/bundler/bundler/issues/5879

morgoth:
@geoffharcourt did you have some time to look into it maybe? :)

geoffharcourt:
Hi @morgoth I hadn't had time recently to do a thorough check. This looks good to me, let's merge and see if this opens things up for other PRs to move forward.

geoffharcourt:
@morgoth can you update the commit message so that it's clear you did? I'll merge after that.

morgoth:
@geoffharcourt Just did and the build is still green

geoffharcourt:
Thank you for your work and your patience!

PR #2496, opened 2017-10-11
Removed extra whitespace from UPGRADING

gfranco93:
I noticed an extra whitespace character in the layout, it looked ugly in the CLI when installing the gem.


geoffharcourt:
Thank you for the fix!

PR #2499, opened 2017-10-18
Update media_type_spoof_detector.rb

vikascoupa:
add a nil check before calling chomp on type_from_file_command


mike-burns:
This indicates a failure elsewhere, in your `file(1)` command. I'd rather not hide that and instead I'd rather be as noisy as possible.

Additionally, having `#calculated_content_type` produce `nil` will wreck havoc on the rest of the methods that depend on it.

PR #2511, opened 2017-11-28
Ignore "gem "bundler" cannot be uninstalled error message

benoittgt:
Refer https://travis-ci.org/rails/rails/jobs/295600391#L1801

This workaround should be removed once https://github.com/bundler/bundler/issues/6072 is addressed.

This is a copy of  : https://github.com/rails/rails/pull/31029

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,8 @@ rvm:
 
 # FIXME: fails with modern bundler
 before_install:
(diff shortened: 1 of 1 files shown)


benoittgt:
Fixed with https://github.com/thoughtbot/paperclip/commit/80847b44f1965d20a2ed6e98a3fa4bf4c1194515

PR #2492, opened 2017-10-03
Fix migration generator for versioned migrations in the Rails 5.

aditya-kapoor:

PR #2521, opened 2017-12-23
Update README.md fix#2465

Maartz:
I've meet this issue 
`StandardError: An error has occurred, this and all later migrations canceled:
Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:
  class AddAttachmentAvatarToUsers < ActiveRecord::Migration[4.2]`

So I've added [5.1] after ActiveRecord::Migration, then I saw in issues it's a known issue, so I've updated the Readme file to prevent for opening issue(s) in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,7 @@ end
 (Or you can use the Rails migration generator: `rails generate paperclip user avatar`)
 
 Add your Rails version after `AddAvatarColumnsToUsers < ActiveRecord::Migration`
+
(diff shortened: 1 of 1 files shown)


morgoth:
@Maartz it's addressed in https://github.com/thoughtbot/paperclip/pull/2492/files and https://github.com/thoughtbot/paperclip/pull/2447/files so when one of it is merged there should be no need for changes in the readme

mike-burns:
Closing in favor of #2447.

PR #2532, opened 2018-01-24
Add Errno::EEXIST as possible exception that happens on linking

hron:
When I try to attach a file on Windows Paperclip fails on trying to make a link with Errno::EEXIST exception.
This exception should be caught as others inside `Paperclip::AbstractAdapter#link_or_copy_file`


mike-burns:
I've merged this without QA because I don't have access to a Windows computer easily, but to confirm: does this work? EEXIST seems like it means cp will also fail?

hron:
Yep, it works for me. I have stumbled upon this problem on a paperclip update and debug the problem. Everything works smoothly when this exceptions is caught.

mike-burns:
Thank you, @hron ! We'll release 5.2.1 shortly to fix this for everyone.

PR #2535, opened 2018-01-28
Increase attachment file size from int (4 bytes) to bigint (8 bytes).

alenzamanyan:
I think the 4 byte limit is unnecessary. Here's a stackoverflow thread that also mentions this problem:
https://stackoverflow.com/questions/34477248/rails-paperclip-rangeerror/47999887#47999887

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,8 @@
 master:
 
+* Change database column type of attachment file size from unsigned 4-byte
+  `integer` to unsigned 8-byte `bigint`. The former type limits attachment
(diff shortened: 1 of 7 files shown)


mike-burns:
This looks like it will only affect new migrations, so there's no backward compatibility concern here. I'll merge this.

mike-burns:
Figured it out: need to check `#sql_type`, not `#type`.

Merged as 34ec355e43e91c63288aab956a604f17471d4e59. Thank you!

PR #2536, opened 2018-01-29
Update the README Installation section

febeling:
The command from the README is often copied into projects directly. The version used here before had a known vulnerability (CVE-2017-0889), so it might be wise to mention something more secure instead.


febeling:
The vulnerability in question can be found here: https://nvd.nist.gov/vuln/detail/CVE-2017-0889

PR #2544, opened 2018-02-06
Do not close files unadapted in post_process_style method

akihikodaki:
Files unadapted while post-processing styles except the original file are closed with `unlink_files`, so it does not have to be removed by `post_process_style`. This eliminates the need of `file.close` in `unlink_files` as well.

The original file should not be closed for the consistency of the input for post-processors.


mike-burns:

PR #2546, opened 2018-02-07
Log failing flush delete actions with logger

martincik:
**As is:**
When deleting attachments (local or remote) and deletion fails all the errors are silently rescued.

**I want:**
When deleting attachments and some fail with an exception I want to see the exception being logged so I can inspect it later when looking at the logs.


mike-burns:

PR #2481, opened 2017-08-31
Added support for aws-sdk-s3 gem which is now preferred way to intera…

morgoth:
…ct with s3.

Reference: https://github.com/aws/aws-sdk-ruby/blob/master/V3_UPGRADING_GUIDE.md#library-maintainer

It still should work with `aws-sdk` fine

PR #2557, opened 2018-02-24
Depends on aws-sdk-s3 gem instead of aws-sdk

iggant:
Gem aws-sdk use about 30 gems as a dependents, but paperclip needs only aws-sdk-s3


morgoth:
already done in https://github.com/thoughtbot/paperclip/pull/2481

Issue #2558, opened 2018-02-25
0 byte file after "Link failed with Invalid cross-device link"


kevgrig:
I have a function that uses Magick to rotate a paperclip-uploaded image and saves the ActiveRecord object with something roughly like this:

```
tfile = # make temporary file
existing_file = # get file bytes from paperclip ActiveRecord file
magick_image = # load existing_file into Magick
magick_image.rotate!(degrees)
magick_image.write(tfile.path)
@obj.file = ActionDispatch::Http::UploadedFile.new(
  tempfile: tfile,
  filename: @obj.file_file_name,
  type: @obj.file_content_type
)
@obj.save!
```
The temporary file is created underneath the rails `tmp` directory and so paperclip tries to do various symlinking and fails:

```
INFO -- : [paperclip] Trying to link /work/myplaceonline/src/myplaceonline/src/myplaceonline_rails/tmp/myp/file6400_20180224-15230-1crpusl to /tmp/090d3fc5d13a0369909a6be48a448b1e20180224-15230-1kiboio.jpeg
INFO -- : [paperclip] Link failed with Invalid cross-device link @ rb_file_s_link - (/work/myplaceonline/src/myplaceonline/src/myplaceonline_rails/tmp/myp/file6400_20180224-15230-1crpusl, /tmp/090d3fc5d13a0369909a6be48a448b1e20180224-15230-1kiboio.jpeg); copying link /work/myplaceonline/src/myplaceonline/src/myplaceonline_rails/tmp/myp/file6400_20180224-15230-1crpusl to /tmp/090d3fc5d13a0369909a6be48a448b1e20180224-15230-1kiboio.jpeg
```

The problem is that [Paperclip::AbstractAdapter.link_or_copy_file](https://github.com/thoughtbot/paperclip/blob/c2c51cab328679ed1b9e4fb49c74ee1b0e652ee0/lib/paperclip/io_adapters/abstract_adapter.rb#L59) doesn't open `@destination` in this error path:

    def link_or_copy_file(src, dest)
      Paperclip.log("Trying to link #{src} to #{dest}")
      FileUtils.ln(src, dest, force: true) # overwrite existing
      @destination.close
      @destination.open.binmode
    rescue Errno::EXDEV, Errno::EPERM, Errno::ENOENT, Errno::EEXIST => e
      Paperclip.log("Link failed with #{e.message}; copying link #{src} to #{dest}")
      FileUtils.cp(src, dest)
    end

[...]

PR #2540, opened 2018-01-31
fix fallback when symlink in AbstractAdapter#link_or_copy_file fails

mraidel:
Paperclip::AbstractAdapter provides the link_or_copy_file method which first tries to do a symlink. When the symlinking raises an exception it is rescued and it tries again with copy. The problem is that the failed symlink leaves the tempfile instance in a bad state and you cannot read from it without reopening it.

Current behavior with a file which cannot be linked (for example because it is on another filesystem):

```
File.open("file_which_cannot_be_linked", "w") { |f| f.print("body") }

Paperclip.io_adapters.for(File.open("file_which_cannot_be_linked")).read
[paperclip] Trying to link file_which_cannot_be_linked to /tmp/d783e2fb96492f5a9f5e9632d2987a1120180131-3949-1vavjso                                                                                                
[paperclip] Link failed with Invalid cross-device link @ rb_file_s_link - (file_which_cannot_be_linked, /tmp/d783e2fb96492f5a9f5e9632d2987a1120180131-3949-1vavjso); copying link file_which_cannot_be_linked to /tm
p/d783e2fb96492f5a9f5e9632d2987a1120180131-3949-1vavjso    
Command :: file -b --mime 'file_which_cannot_be_linked'                              
=> ""
```
Behavior after fix:

```
Paperclip.io_adapters.for(File.open("file_which_cannot_be_linked")).read                                                                                                                  
[paperclip] Trying to link file_which_cannot_be_linked to /tmp/d783e2fb96492f5a9f5e9632d2987a1120180131-3939-1cp0zud                                                                                              
[paperclip] Link failed with Invalid cross-device link; copying link file_which_cannot_be_linked to /tmp/d783e2fb96492f5a9f5e9632d2987a1120180131-3939-1cp0zud                                         
Command :: file -b --mime 'file_which_cannot_be_linked'                                                                                                                                                             
=> "body"                                                                                                                                                                                        
```

PR #2559, opened 2018-02-25
Issue #2558: Also open destination file in symlink failure case

kevgrig:
See Issue #2558

### DIFF
--- a/lib/paperclip/io_adapters/abstract_adapter.rb
+++ b/lib/paperclip/io_adapters/abstract_adapter.rb
@@ -64,6 +64,7 @@ module Paperclip
     rescue Errno::EXDEV, Errno::EPERM, Errno::ENOENT, Errno::EEXIST => e
       Paperclip.log("Link failed with #{e.message}; copying link #{src} to #{dest}")
       FileUtils.cp(src, dest)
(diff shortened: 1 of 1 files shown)


garettarrowood:
This is a duplicate resolution to #2540 . Nevertheless - without this fix, a `paperclip` security upgrade turned into a major bug for my work repo. I've since forked and applied this exact change, and it fixed our problem. Context: All of our call center recordings failed to upload with no reasonable looking error output. This caused Argument errors in model creation database transactions. Nothing pointed specifically to `[paperclip]`.

TLDR; `paperclip` version 5.2.1 broke our application.  

There's a lot of activity going on the repo, but I recommend merging this and cutting another patch release to resolve this hard to diagnose bug. Would be nice to get the Ruby 2.5.0 `Fixnum` deprecation warnings resolved too, which has already been merged but is unreleased :). 

Thanks for this fix! And _owners_, thanks for maintaining an awesome tool!

cc/ @sidraval @mike-burns @cbeckr

sidraval:
Thanks @kevgrig & @garettarrowood !  See https://github.com/thoughtbot/paperclip/pull/2567

PR #2560, opened 2018-02-26
update to use new aws-sdk version 3 structure

jdguzman:
The AWS SDK now has separate gems for each resource this changes paperclip to only use what it needs.

This change only requires the use of the `aws-sdk-s3` gem. I've updated the readme also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -567,7 +567,7 @@ Storage
 Paperclip ships with 3 storage adapters:
 
 * File Storage
-* S3 Storage (via `aws-sdk`)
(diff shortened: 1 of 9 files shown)


pjg:
There was already a very similar commit in the past, which I cannot locate now apart from its sha1: https://github.com/thoughtbot/paperclip/commit/d53818356375fc2fb5f341ee81354afa7fab012f

morgoth:
@jdguzman This is already done and released in version 6.0.0

PR #2568, opened 2018-03-02
A migration guide from Paperclip to ActiveStorage

mike-burns:
A start at an introductory migration path for moving from Paperclip to
[ActiveStorage].

ActiveStorage is coming in Rails 5.2 (April, perhaps) and we want to
help people move from Paperclip to it. Having a file upload strategy
baked into Rails itself will allow us all to focus our efforts and
follow a common goal.

This migration guide is rudimentary at best: it works for a `User`
object with one `avatar` attachment. It is provided as a framework to
start the documentation.

I have tried this in a dummy app.

[ActiveStorage]: http://edgeguides.rubyonrails.org/active_storage_overview.html

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -1,0 +1,278 @@
+# Migrating from Paperclip to ActiveStorage
+
+Paperclip and ActiveStorage solve similar problems with similar solutions, so
+transitioning from one to the other is straightforward data re-writing.
+
+The process of going from Paperclip to ActiveStorage is as follows:
+
+1. Apply the ActiveStorage database migrations.
+2. Configure storage.
+3. Copy the database data over.
+4. Copy the files over.
+5. Update your tests.
+6. Update your views.
+7. Update your controllers.
+8. Update your models.
+
+## Apply the ActiveStorage database migrations
+
+Follow [the instructions for installing ActiveStorage]. You'll very likely want
+to add the `mini_magick` gem to your Gemfile.
+
+```sh
+rails active_storage:install
+```
+
+[the instructions for installing ActiveStorage]: https://github.com/rails/rails/blob/master/activestorage/README.md#installation
+
+## Configure storage
+
+Again, follow [the instructions for configuring ActiveStorage].
+
+[the instructions for configuring ActiveStorage]: http://edgeguides.rubyonrails.org/active_storage_overview.html#setup
+
+## Copy the database data over
+
+The `active_storage_blobs` and `active_storage_attachments` tables are where
+ActiveStorage expects to find file metadata. Paperclip stores the file metadata
+directly on the associated object's table.
+
+You'll need to write a migration for this conversion. Because the models for
+your domain are involved, it's tricky to supply a simple script. But we'll try!
+
+Here's how it would go for a `User` with an `avatar`, that is this in
+Paperclip:
+
+```ruby
+class User < ApplicationRecord
+  has_attached_file :avatar
+end
+```
+
+Your Paperclip migrations will produce a table like so:
+
+```ruby
+create_table "users", force: :cascade do |t|
+  t.string "avatar_file_name"
+  t.string "avatar_content_type"
+  t.integer "avatar_file_size"
+  t.datetime "avatar_updated_at"
+end
+```
+
+And you'll be converting into these tables:
+
+```ruby
+create_table "active_storage_attachments", force: :cascade do |t|
+  t.string "name", null: false
+  t.string "record_type", null: false
+  t.integer "record_id", null: false
+  t.integer "blob_id", null: false
+  t.datetime "created_at", null: false
+  t.index ["blob_id"], name: "index_active_storage_attachments_on_blob_id"
+  t.index ["record_type", "record_id", "name", "blob_id"], name: "index_active_storage_attachments_uniqueness", unique: true
+end
+```
+
+```ruby
+create_table "active_storage_blobs", force: :cascade do |t|
+  t.string "key", null: false
+  t.string "filename", null: false
+  t.string "content_type"
+  t.text "metadata"
+  t.bigint "byte_size", null: false
+  t.string "checksum", null: false
+  t.datetime "created_at", null: false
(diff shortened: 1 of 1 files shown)


mike-burns:
Ping @geoffharcourt -- thanks for your support when I started this document!

MIGRATING.md
```diff
@@ -0,0 +1,278 @@
...
+    # instance.send("#{attachment}_file_name")
+  end
+
+  def checksum(attachment)
+    # local
+    url = attachment.path
+    # remote
+    # url = attachment.url
+
+    Digest::MD5.base64digest(open(url).read)
```
derekprior: This is probably ok since theoretically you had good control over the URL that was written to your database, but `Kernel.open` makes me really [nervous](https://sakurity.com/blog/2015/02/28/openuri.html). So I just wanted to share my nerves so you could probably assuage them. I'm not sure of any specific attack vector here, but OpenURI can really do damage.

MIGRATING.md
```diff
@@ -0,0 +1,278 @@
...
+In Paperclip it looks like this:
+
+```ruby
+image_tag @user.avatar.url(:medium)
+```
+
+In ActiveStorage it looks like this:
+
+```ruby
+image_tag @user.avatar.variant(resize: "250x250")
```
derekprior: is it worth calling out that this may require preloading `avatar`? My understanding is that paperclip has the necessary columns for `avatar` on its `User` model, while the major difference with ActiveStorage is that it's a polymorphic relationship? Is this correct? Out of scope for this doc, maybe?

MIGRATING.md
mike-burns: No, you're right. Maybe this method would be safer either explicitly expecting a file path or explicitly expected a HTTP URL.

If it's a file path, use `File.read` instead of `open(url).read`.

If it's a URL use `Net::HTTP.get(URI(url).host, URI(url).path)`.

Does that seem right, or should the Net::HTTP code to be fancier?

MIGRATING.md
mike-burns: Both a fair point and out of scope for this doc, but maybe it's worth mentioning in this document because we will need to watch for that when converting.

So `@user = User.find(params[:id])` becomes `@user = User.find(params[:id]).include(:funny_pictures)`, for a `has_many_attached :funny_pictures`. That's where we'll run into the most problems, right -- a has-many relationship?

derekprior: I was also thinking about his particular case (for example, registering the avatar of each commenter in a list of comments on a post). Maybe a heads up on this could go in the currently-empty controllers section?

MIGRATING.md
derekprior: I never know how to do anything in `Net:HTTP`, but yeah. I think that approach is good.

MIGRATING.md
```diff
@@ -0,0 +1,312 @@
...
+Instead of the `have_attached_file` matcher, you'll need to write your own.
+Here's one that is similar in spirit to the Paperclip-supplied matcher:
+
+```ruby
+RSpec::Matchers.define :have_attached_file do |name|
+  matches do |record|
+    file = record.send(name)
+    file.respond_to?(:variant) && file.respond_to?(:attach)
+  end
+end
```
geoffharcourt: You went the extra mile with this matcher.

sankage:
I just tried running the migration using `pg (1.0.0)` and `psql (PostgreSQL) 10.3` and get a bunch of errors from prepared statements.

1. First error: 

   ```
   ArgumentError: wrong number of arguments (given 1, expected 2..3)
   [...]_convert_to_active_storage.rb:14:in `prepare'    
   ```

   I added names to the prepared statements:

   ```ruby
   ActiveRecord::Base.connection.raw_connection.prepare('active_storage_blob_statement', <<-SQL)

   ActiveRecord::Base.connection.raw_connection.prepare('active_storage_attachment_statement', <<-SQL)
   ```

2. And now the error is:

   ```
   ERROR:  syntax error at or near ","
   LINE 4:       ) VALUES (?, ?, ?, '{}', ?, ?, ?)
                            ^
   [...]_convert_to_active_storage.rb:14:in `prepare'
   ```

   I added numbered placeholders:

   ```
   ) VALUES ($1, $2, $3, '{}', $4, $5, $6)

   ) VALUES ($1, $2, $3, #{get_blob_id}, $4)
   ```

3. I then get the error:

   ```
   NoMethodError: undefined method `execute' for #<PG::Result>                                                                                
   [...]_convert_to_active_storage.rb:39:in `block (4 levels) in up'   
   ```

   I switch the `execute` command with `exec_prepared`:

   ```ruby
   ActiveRecord::Base.connection.raw_connection.
     exec_prepared("active_storage_blob_statement", [
       key(instance, attachment),
       instance.send("#{attachment}_file_name"),
       instance.send("#{attachment}_content_type"),
       instance.send("#{attachment}_file_size"),
       checksum(instance.send(attachment)),
       instance.updated_at.iso8601
     ])

   ActiveRecord::Base.connection.raw_connection.
     exec_prepared("active_storage_attachment_statement", [
       attachment, model.name, instance.id, instance.updated_at.iso8601
     ])
   ```

4. Now the error:

   ```
   NoMethodError: undefined method `close' for #<PG::Result>
   [...]_convert_to_active_storage.rb:60:in `up'  
   ```

   Fixed by removing the 2 `close` calls:

   ```ruby
   active_storage_attachment_statement.close
   active_storage_blob_statement.close
   ```


After making these changes I was able to get the migration to complete.

mike-burns:
Wonderful report, thank you @sankage ! I'll look into this on Friday.

mike-burns:
@sidraval - I won't have time this coming Friday, but if you do: consider the above when looking at #2596.

GBH:
Any reason that Github issues got removed? I'm sure that there are many users that still need access there to debug their existing installs.

mike-burns:
Thanks @sankage , I've opened a PR: https://github.com/thoughtbot/paperclip/pull/2613 .

PR #2572, opened 2018-03-08
Rewind file in Storage::Fog if bucket does not exist

natesholland:
There was a bug in the code where if you created a bucket the file would
get read but not rewind on the first pass through (when the bucket
didn't exist). So when we went into the retry logic the file would
already be read and zero bytes would get copied over from the file. This
fixes that by rewinding the file during the retry so it gets correctly
copied over the second time around.

### DIFF
--- a/lib/paperclip/storage/fog.rb
+++ b/lib/paperclip/storage/fog.rb
@@ -117,6 +117,7 @@ module Paperclip
             raise if retried
             retried = true
             directory.save
+            file.rewind
             retry
           ensure
             file.rewind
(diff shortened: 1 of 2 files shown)


sidraval:
Thanks for the PR 👍

natesholland:
I know this is a stretch @sidraval but is there any chance you could port this over into the `v4.3` branch? We are currently bumping up to 4.3.7 and had to monkey patch around this issue (hence the PR). It would be greatly appreciated if thoughtbot could port this fix over into the 4.3 branch.

sidraval:
@natesholland I've got it on a branch: https://github.com/thoughtbot/paperclip/tree/v4.3.8 but I'm experiencing CI failures that I don't have the bandwidth to deal with at the moment. You're welcome to take a peek, but I don't think I should cut v4.3.8 without passing CI at least.

PR #2573, opened 2018-03-08
Rewind file in Storage::Fog if bucket does not exist

natesholland:
There was a bug in the code where if you created a bucket the file would
get read but not rewind on the first pass through (when the bucket
didn't exist). So when we went into the retry logic the file would
already be read and zero bytes would get copied over from the file. This
fixes that by rewinding the file during the retry so it gets correctly
copied over the second time around.

This is the same patch as #2572 but into the `v4.3` branch. My project is
currently on 4.3.7 so I would appreciate the port down into the `v4.3` branch
as well as into master.

### DIFF
--- a/lib/paperclip/storage/fog.rb
+++ b/lib/paperclip/storage/fog.rb
@@ -114,6 +114,7 @@ module Paperclip
             raise if retried
             retried = true
             directory.save
+            file.rewind
             retry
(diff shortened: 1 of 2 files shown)


sidraval:
Closing as a duplicate #2572

Issue #2588, opened 2018-04-11
Rollback after save results in the attachment pointing to a deleted file


kmewhort:
For example:
```
# save an initial file
record.file = File.new('my_file_a.jpg')
record.save

# save a second replacement file
ActiveRecord::Base.transaction do
  record.file = File.new('my_file_b.jpg')
  record.save
  raise ActiveRecord::Rollback
end
```
At the end of this, the attachment file name will be pointing to the original `my_file_a.jpg`, but this has been deleted.

A workaround I'm currently using is the following:
```
has_attached_file :file, keep_old_files: true

after_commit on: [:update, :destroy] do
  self.send(document_name).flush_deletes
end
```

This generally works....but does create issues if you destroy and then re-save a file with an identical filename.  For example:
```
record.file = File.new('test_a/identical_filename.jpg')
record.save

ActiveRecord::Base.transaction do
  record.file = nil
  record.file = File.new('test_b/identical_filename.jpg')
end
```
The file is queued for deletion on commit, then the new file is saved....then the new file is deleted.  A proper fix should probably move the old file to a temporary file first (so it's not clobbered), then queue that one for deletion.

For now, as a workaround, I'm just ensuring filenames are unique:
```
attachment_options[:filename_cleaner] = Proc.new do |filename|
  default_cleaner = Paperclip::FilenameCleaner.new(Paperclip::Attachment.default_options[:restricted_characters])
  filename = default_cleaner.call(filename)

  extension = File.extname(filename).downcase
  base = File.basename(filename, extension)
  base.gsub!(/\-x[\da-f]{6}$/, '')
  hash = Digest::SHA1.hexdigest(Time.current.to_s).slice(0..5)
  "#{base}-x#{hash}#{extension}"
end
```

sidraval:
Thank you for reporting this. Unfortunately, we will be deprecating Paperclip and therefore will not have the bandwidth to address this issue.

PR #2589, opened 2018-04-11
Support db transaction rollbacks with option to keep files until commit

kmewhort:
This PR:
  1) adds a `:keep_files_until_commit` option to support database transaction rollbacks, and
  2) fixes an issue with where a new file with the same filename is deleted on commit if using this option (the issue actually already exists if using the `:keep_files` option).

Further details in the issue here:  https://github.com/thoughtbot/paperclip/issues/2588.

I'll add tests and specs, but would appreciate feedback on the approach first, thanks!

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -112,6 +112,10 @@ module Paperclip
           assign_attributes
           post_process_file
           reset_file_if_original_reprocessed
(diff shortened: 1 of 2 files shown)


mike-burns:

(This is a pretty cool idea that would have helped me a bunch, too.)

PR #2600, opened 2018-05-14
Add Spanish translation to MIGRATING

HoracioChavez:


mike-burns:
1. Wow. Amazing. I love it.
2. Now we face the classic translators' problem: what do we do when the source text is updated? What do you think, @HoracioChavez ?

HoracioChavez:
1. Thanks 😊
2. You’re right, updating the source text will be necessary. I would be happy to maintain the translation, if that’s okay?

mike-burns:
Thank you again! I've merged this as 3f1512fdc4940ef6b97e5c25128612e37d871ee9, and linked to it from the README.

PR #2602, opened 2018-05-17
Fix bug in migration instructions on MariaDB

banjerluke:
"key" is a reserved keyword in MariaDB, so the SQL statement fails. Surrounding the keyword in backticks fixes the error.


mike-burns:
Thank you! Merged as 9e40a5d5ac8ea8e9a80f073faabc78e5490f4ebd.

PR #2616, opened 2018-10-05
Expand upon Migration Documentation

seanpdoyle:
Explain some lessons learned from successfully migrating an application
from Paperclip to ActiveStorage.

Replace links to `master` or `edgeguides.rubyonrails.org` with versions
locked to `v5.2`.

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -16,20 +16,49 @@ The process of going from Paperclip to ActiveStorage is as follows:
 
 ## Apply the ActiveStorage database migrations
 
-Follow [the instructions for installing ActiveStorage]. You'll very likely want
-to add the `mini_magick` gem to your Gemfile.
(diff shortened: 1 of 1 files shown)


MIGRATING.md
```diff
@@ -344,16 +392,35 @@ And your view has
...
+attachments and blobs.
+
+A `has_many_attached :avatars` declaration will generate a `has_many
+:avatar_attachments` relationship along with a
+[`.with_attached_avatars`][has-many-eager-loading-scope] scope for eager loading
+attachments and blobs.
+
+When eager-loading transitive relationships, you'll need to specify the
+relationship names directly, like `includes(avatar_attachment: :blob)` or
+`includes`(avatar_attachments: :blob)`:
```
composerinteralia: I think there is an extra backtick here:

```
`includes`(avatar_attachments: :blob)`
```

PR #2617, opened 2018-11-01
Fix a couple of typos in the migration guide

paulmthiebauth:
The Paperclip -> ActiveStorage migration guide was referencing the recommended validation gem incorrectly so I've updated it

Additionally, there were backticks around the `key` value in the active_storage_blobs SQL that were causing the migration to error out when trying to run it

This PR addresses both of those issues

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -134,7 +134,7 @@ class ConvertToActiveStorage < ActiveRecord::Migration[5.2]
 
     active_storage_blob_statement = ActiveRecord::Base.connection.raw_connection.prepare('active_storage_blob_statement', <<-SQL)
       INSERT INTO active_storage_blobs (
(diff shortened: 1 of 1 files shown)


mike-burns:
Whoops, I merged when I should have asked questions.

First off, thank you! I've kept the file-validation changes.

`key` is a reserved word in (some) SQL. It looks like there's some confusion on the Web around how to escape these -- I had always used backticks, but apparently it's the SQL standard to use double quotes. But then MySQL needs a flag if you want to use double quotes instead of backticks.

Not totally sure what to do here. I've put the backticks back for now.

What DBMS did you run into issues on?

paulmthiebauth:
@mike-burns Sorry for the delay, I'm using PostgreSQL locally

PR #2619, opened 2018-11-09
Avoid dynamic extend

printercu:
Hi!

I see this gem is deprecated. However maybe this changes will be suitable for the last release.

It removes calls to `Object#extend`  every time attachment is created. This makes:
- storage initialization code run only once, when first attachment with this storage is instantiated,
- use ruby's `super` instead of custom `initialize_storage` hooks,
- avoid bumping class_serial, which flushes ruby's method cache.

### DIFF
--- a/lib/paperclip/attachment.rb
+++ b/lib/paperclip/attachment.rb
@@ -37,6 +37,27 @@ module Paperclip
       }
     end
 
+    def self.attachment_class_cache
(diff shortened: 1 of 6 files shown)


mike-burns:
Hi, thank you for this. It looks useful.

But unfortunately we are no longer taking code changes.

PR #2629, opened 2018-12-21
Add fog_options as an extra param to expiring_url

dgilperez:
Fog's [get_https_url](https://www.rubydoc.info/github/fog/fog-aws/Fog%2FStorage%2FAWS%2FFiles:get_https_url) accepts an optional `options` hash parameter, that is used to provide extra keys to the url, such as for example requesting some particular header values in the response. 

This PR allows for Paperclip's `expiring_url` to accept the same optional `fog_options` attribute, that will be directly passed to `get_https_url`. 

This also makes the `Fog` and `S3` `Storage` modules more homogenous, since S3 was already accepting `s3_url_options` for the same purpose on its own `expiring_url`.

### DIFF
--- a/lib/paperclip/storage/fog.rb
+++ b/lib/paperclip/storage/fog.rb
@@ -149,11 +149,11 @@ module Paperclip
         end
       end
 
-      def expiring_url(time = (Time.now + 3600), style_name = default_style)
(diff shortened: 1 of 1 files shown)


mike-burns:
Thank you for this great submission. Unfortunately, since Paperclip is deprecated, we are not accepting any code changes.

PR #2625, opened 2018-12-01
Add missing "not" in Migration Guide Validations section

TRMW:
I believe this is meant to be "does not", not "does".

PR #2630, opened 2019-01-31
Build in validation

eternalfool:
ActiveStorage does _not_ have built-in support for Validation


mike-burns:
Thank you for catching this! Merged via #2625 .

PR #2636, opened 2019-03-24
FIX: warning instance variable @attachment_ not initialized

bjelline:
check existance of instance variable before using it.
this problem was mentioned in #1432


mike-burns:
Great fix, thank you! However, since Paperclip is deprecated, we are no longer accepting code changes.

PR #2637, opened 2019-04-12
Critical Bug Fix: Use AWS::S3::Object built in download_file method to avoid truncating during #reprocess!

daniel-nelson:
Yesterday, I ran `#reprocess!` and it truncated all original images on S3. [The logs](https://gist.github.com/daniel-nelson/c865bc5e3bcdd547ee2841e0dd660b23) show that there was an error processing the image, and then the downloaded image was uploaded again.

I inspected the temporary file created from the download, and it was empty, so that explains why the exceptions were thrown and why the original was truncated (Paperclip uploaded the empty image back to S3).

Switching to the AWS::S3::Object built in `download_file` method fixes this.

### DIFF
--- a/lib/paperclip/storage/s3.rb
+++ b/lib/paperclip/storage/s3.rb
@@ -412,11 +412,7 @@ module Paperclip
 
       def copy_to_local_file(style, local_dest_path)
         log("copying #{path(style)} to local file #{local_dest_path}")
-        ::File.open(local_dest_path, 'wb') do |local_file|
-          s3_object(style).get do |chunk|
-            local_file.write(chunk)
-          end
-        end
+        s3_object(style).download_file(local_dest_path)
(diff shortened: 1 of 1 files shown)


mike-burns:
Thank you for catching this and the great fix. However, since Paperclip is deprecated, we are no longer accepting code changes.

aka47:
someone should take over paperclip then .. at least for bug fixes.

waissbluth:
for anyone arriving here from the paperclip docs: https://github.com/kreeti/kt-paperclip

sergey-alekseev:
To clarify the previous comment – the fix was merged in https://github.com/kreeti/kt-paperclip/pull/9 and released in https://github.com/kreeti/kt-paperclip/releases/tag/v6.2.0, so you can switch to https://rubygems.org/gems/kt-paperclip.

PR #2638, opened 2019-04-19
Fixes annoying warnings in filesystem.rb

fun-ruby:
# paperclip-4.3.7/lib/paperclip/storage/filesystem.rb:49: warning: `&' after local variable or literal is interpreted as binary operator
# paperclip-4.3.7/lib/paperclip/storage/filesystem.rb:49: warning: even though it seems like argument prefix


mike-burns:
Paperclip is deprecated and no longer accepting code changes. Thank you!

PR #2654, opened 2020-01-24
update readme, use kt-paperclip for cucumber tests and migration

shalinibhawsingka-kreeti:


shalinibhawsingka-kreeti:
This is an invalid PR

PR #2655, opened 2020-02-14
Link to kt-paperclip

mike-burns:
We were contacted in early December by Surendra of Kreeti, who explained
that they had forked Paperclip and were maintaining it. We let the
December US holidays go by and then checked in; they're still
maintaining it.

We do not know any of the Kreeti devs and cannot vouch for them as a
team. We did not know of a way to hand over the Rubygem ownership while
expressing that people should remember to audit when upgrading.
Therefore, we felt it best that they have a full fork: new Rubygem gem,
separate repo.

Therefore it makes sense for us to link to Kreeti's fork as an option
for those who cannot move off of Paperclip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,12 @@ For existing projects, please consult and contribute to the migration guide,
 available [in English], [en español], and as [a video] recorded at RailsConf
 2019.
 
+Alternatively, for existing projects, [Kreeti] (insecure link) is maintaining
+[kt-paperclip], an ongoing [fork of Paperclip].
 
 We will leave the Issues open as a discussion forum _only_. We do _not_
-guarantee a response from us in the Issues.
+guarantee a response from us in the Issues. All bug reports should go to
+kt-paperclip.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,6 +26,9 @@ migration guide. All other pull requests will be closed without merging.
 [in English]: https://github.com/thoughtbot/paperclip/blob/master/MIGRATING.md
 [en español]: https://github.com/thoughtbot/paperclip/blob/master/MIGRATING-ES.md
 [a video]: https://www.youtube.com/watch?v=tZ_WNUytO9o
+[Kreeti]: https://www.kreeti.com/
```
thorncp: My browser gives me a cert error when I follow this link. Should we drop the `s`?

mike-burns: Yeah I guess so. I've noted in the text that it's a cleartext link just in case someone has an ISP that screws with that, or is otherwise paranoid.

PR #2661, opened 2020-06-21
Fix Syntax Error

mzagaja:
When running the migration as typed a minor syntax error causes the migration to fail. The [linked video](https://youtu.be/tZ_WNUytO9o?t=617) shows the correct syntax.


rhombl4:
Perhaps it requires more complicated fixing.

PR #2672, opened 2021-02-23
fix calling open-uri

manuelmeurer:
I upgraded from Rails 6.0 to 6.1, and now calling open-uri via `Kernel#open` doesn't seem to recognize `URI` objects anymore:

`TypeError: no implicit conversion of URI::HTTPS into String`

Calling it via `URI.open` fixes this.


manuelmeurer:
I found the fork by Kreeti and saw that this issue is fixed there:
https://github.com/kreeti/kt-paperclip/pull/45

PR #2675, opened 2021-03-01
Use CGI.escape and CGI.unescape instead of URI.escape and URI.unescape

sushantmittal:
Used `CGI.escape` and `CGI.unescape` as `URI.escape` and `URI.unescape` became obsolete and incompatible with ruby 3.0.0.


sasharevzin:
Looks good. Very sad that this repo archived. May be you can create a néw PR at it’s fork? https://github.com/kreeti/kt-paperclip

PR #2676, opened 2021-03-14
Update url_generator.rb

iqre8:
undefined method `escape' for URI:Module with Ruby 3.0.0
This method is obsolete and should not be used. Instead, use
CGI.escape, URI.encode_www_form or URI.encode_www_form_component
depending on your specific use case.


sasharevzin:
Looks good to me. Would you submit it to https://github.com/kreeti/kt-paperclip

asecondwill:
is kt-paperclip a direct replacement?  is paperclip completely unsupported now?

PR #2677, opened 2021-03-24
Remove mimemagic as it now released under a GPL license, incompatible with Paperclip's MIT license

sd:
NOTE: this is still untested work in progress. 

Opened this PR in case anyone wants to help me.

See https://github.com/rails/rails/issues/41750 for more info.

Issue https://github.com/thoughtbot/paperclip/issues/2678

NOTE: Since paperclip has mandatory spoof detection, you will need to have the `file` command installed on your server in order for file uploads to keep working.

### DIFF
--- a/lib/paperclip/content_type_detector.rb
+++ b/lib/paperclip/content_type_detector.rb
@@ -60,16 +60,10 @@ module Paperclip
     end
 
     def type_from_file_contents
-      type_from_mime_magic || type_from_file_command
+      type_from_file_command
     rescue Errno::ENOENT => e
       Paperclip.log("Error while determining content type: #{e}")
       SENSIBLE_DEFAULT
-    end
-
-    def type_from_mime_magic
-      @type_from_mime_magic ||= File.open(@filepath) do |file|
-        MimeMagic.by_magic(file).try(:type)
-      end
     end
 
     def type_from_file_command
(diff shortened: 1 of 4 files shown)


ghiculescu:
might be worth doing this on https://github.com/kreeti/kt-paperclip too

alexshenia:
mimemagic returned back MIT license, I do not think that we need this changes anymore

mathieujobin:
Actually, I am using this branch... I like to remove dependencies when they are not needed.
Could this be flipped into an optional dependency? I don't know if there are people who prefer to use mimemagic for some reason.
My vote would be to merge this

thanks

lib/paperclip/content_type_detector.rb
```diff
@@ -60,18 +60,12 @@ def possible_types
...
-      type_from_mime_magic || type_from_file_command
+      type_from_file_command
     rescue Errno::ENOENT => e
       Paperclip.log("Error while determining content type: #{e}")
       SENSIBLE_DEFAULT
     end
 
-    def type_from_mime_magic
-      @type_from_mime_magic ||= File.open(@filepath) do |file|
-        MimeMagic.by_magic(file).try(:type)
```
mathieujobin, line 71: Or replace with Marcel ? which comes with ActiveStorage

PR #2680, opened 2021-03-25
Removed mimemagic and falling back to file due to incompatible licence

anathematic:
For those who like to live on the bleeding edge, I've updated a 5~ year old version with the mimemagic fix 😎

### DIFF
--- a/lib/paperclip/version.rb
+++ b/lib/paperclip/version.rb
@@ -1,3 +1,3 @@
 module Paperclip
-  VERSION = "4.3.7".freeze unless defined? Paperclip::VERSION
+  VERSION = "4.3.6".freeze unless defined? Paperclip::VERSION
 end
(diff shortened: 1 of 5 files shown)


lib/paperclip/version.rb
```diff
@@ -1,3 +1,3 @@
 module Paperclip
-  VERSION = "4.3.7".freeze unless defined? Paperclip::VERSION
+  VERSION = "4.3.6".freeze unless defined? Paperclip::VERSION
```
theianjones, line 2: isnt the next version `4.3.8`?

```suggestion
  VERSION = "4.3.8".freeze unless defined? Paperclip::VERSION
```

anathematic: Hah! It turns out I can't increment numbers- must have missed that day in school.

I no longer have access to the original repo I made this PR from, so please feel free to close this PR.

Issue #2678, opened 2021-03-24
Paperclip mimemagic dependency licensing implication


sb8244:
Paperclip is deprecated, but there may still be users of the gem that haven't upgraded for whatever reason.

There was a mimemagic change (https://github.com/rails/rails/issues/41750 https://github.com/minad/mimemagic/issues/97) that stems from incorrectly licensing mimemagic as MIT with a GPLv2 dependency. The old versions were yanked (new 0.3.6 added) and new versions are published as GPLv2. Codebases that pull in paperclip will possibly be taking the risk of GPLv2 code (IANAL) instead of MIT code.

I am not sure there's any alternative at the moment, but wanted to create this issue in case someone else runs into it.

## Deprecation notice

Paperclip is currently undergoing [deprecation in favor of ActiveStorage](https://github.com/thoughtbot/paperclip/blob/master/MIGRATING.md). Maintainers of this repository will no longer be tending to new issues. We're leaving the issues page open so Paperclip users can still see & search through old issues, and continue existing discussions if they wish.

sd:
I have a Pull Request that removes the dependency on mimemagic.

https://github.com/thoughtbot/paperclip/pull/2677

Supernats:
https://github.com/kreeti/kt-paperclip is the listed maintained fork. There have not been changes to this repository in over a year.

wisetara:
@sd I hope you'll submit your PR to https://github.com/kreeti/kt-paperclip as @Supernats suggests. I'd rather switch to the "officially" maintained gem than point to your branch, which does not negate my appreciation for your efforts!

sd:
https://github.com/kreeti/kt-paperclip/pull/52

wennerma:
I see there's a few PRs, is there an update on when these could be merged?  Is a better route for some of us to fork this and make the change ourselves?

ryantk:
Has anyone got feedback for @sd 's branch? 
I see the build has failed on Travis but I don't know if this is normal for this project (one would hope not but who knows).
I would like to just use that branch and have done with it but just looking to see if anyone has had any glaring issues.

GProst:
@ryantk FWIW, tested in our project, image creation/update/deletion - looks good
Ruby 2.5.7

sd:
@ryantk we have deployed to prod without issues. We do have the `file` command available on the server. Otherwise Paperclip blocks all uploads due to "anti-spoofing" checks that cannot confirm the file contents.

[...]

PR #2683, opened 2021-03-26
Update mimemagic dependency to 0.3.9 for MIT license

psychezone:
Per Issue #2678, mimemagic dependency 0.3.0 is now yanked due to licensing issue.  However, there is a later 0.3.9 that now has lowered nokogiri and rake dependencies on MIT license.


DmytroStepaniuk:
README.md says 

```
We are no longer accepting pull requests except pull requests against the migration guide. All other pull requests will be closed without merging.
```
😔

PR #2694, opened 2021-07-02
Delete NEWS

Arya128-ux:


mathieujobin (changes requested):
why deleting the readme ? what is file1 ?

